PR #2, opened 2022-05-18
Measure upload

natjoe4:
# Summary
Added a file upload component that accepts .json files for measure upload

## New Behavior
The home screen now contains a file dropzone which will populate a global measure bundle recoil state on file drop

## Code Changes
- Added file dropzone component and testing
- Added some of Matt's recoil testing helper functions

# Testing Guidance
- Run `npm install` then `npm run check`
- Run `npm run dev`
- Click on the file dropzone and upload any .json file. Ensure the file icon changes as well as the text prompt to reflect the uploaded file name
- Add `maxSize={1}` on line 42 of `components/MeasureUpload.tsx`
- Now upload any .json file and endure an error notification appears with a helpful error message

### DIFF
--- a/components/MeasureUpload.tsx
+++ b/components/MeasureUpload.tsx
@@ -1,0 +1,46 @@
+import { Dropzone } from '@mantine/dropzone';
+import { showNotification } from '@mantine/notifications';
+import { Grid, Center, Text } from '@mantine/core';
+import { IconFileImport, IconFileCheck, IconAlertCircle } from '@tabler/icons';
+import { useSetRecoilState, useRecoilValue } from 'recoil';
+import { measureBundleState } from '../state/atoms/measureBundle';
+
+function dropzoneChildren() {
+  const measureBundle = useRecoilValue(measureBundleState);
+  return (
+    <Grid justify="center">
+      <Grid.Col span={12}>
+        <Center>{measureBundle ? <IconFileCheck size={80} /> : <IconFileImport size={80} />}</Center>
+      </Grid.Col>
+      <Grid.Col>
+        <Center>
+          <Text size="xl" inline>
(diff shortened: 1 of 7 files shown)


hossenlopp (changes requested):
This could use further work to verify that the file uploaded is indeed a Measure bundle.

state/atoms/measureBundle.ts
```diff
@@ -3,7 +3,10 @@ import { atom } from 'recoil';
 /**
  * Atom tracking and controlling the value of uploaded measure bundle
  */
-export const measureBundleState = atom<File | null>({
+export const measureBundleState = atom<{ name: string; content: fhir4.Bundle | null }>({
```
hossenlopp, line 6: Could move the `{ name: string; content: fhir4.Bundle }` to it's own interface to neaten this up a bit.

components/MeasureUpload.tsx
```diff
@@ -7,9 +7,40 @@ import { measureBundleState } from '../state/atoms/measureBundle';
...
+    const reader = new FileReader();
+    reader.onload = () => {
+      const bundle = JSON.parse(reader.result as string) as fhir4.Bundle;
+      const numMeasures = bundle?.entry?.filter(r => r?.resource?.resourceType === 'Measure')?.length;
+      if (bundle.resourceType !== 'Bundle') {
+        showNotification({
+          id: 'failed-upload',
+          icon: <IconAlertCircle />,
+          title: 'File upload failed',
+          message: `Uploaded file must contain a resource of type 'Bundle'`,
```
hossenlopp, line 21:
```suggestion
          message: `Uploaded file must be a JSON FHIR Resource of type 'Bundle'`,
```

hossenlopp (approved):
Looks good to me! The recoil stuff will also let us make this drop thing go away when the user is ready to go and we can use the name of the file to indicate what bundle is being worked with in some other component. I like it!

PR #4, opened 2022-05-27
Vs patch to use regex

natjoe4:
# Summary
Updated our method of identifying VSAC valuesets to use Regex

## New Behavior
Valuesets under `path: code` will now be included in required valuesets when checking if all required valuesets are included

## Code Changes
- Use a regex to check if a valueset is a VSAC valueset in `identifyMissingValuesets()` function

# Testing Guidance
- Try dropping a measure bundle missing a required valueset in the file dropzone and ensure you get a popup modal with a link to the missing valueset


mgramigna (approved):
perfect

PR #6, opened 2022-06-02
Patient creation

mgramigna:
# Summary

Adds a button/simple code editor for creating a patient resource. Created patients will show up in the app and are editable/deletable

## New Behavior

New button that triggers patient creation where user can edit the FHIR Patient resource content. Currently generates a random name, and DOB that makes them 21 years old w/r/t the start of the measurement period.

## Code Changes

* Add component for generic codemirror modal. Can be re-used for editing any JSON resource, not just patients
* Add patient creation component that handles logic of instantiating a specific test case
* Add helper code for generating random attributes on the resource, as well as the resource itself (uses same fake name/gender data set as Synthea)
* Use Immer to control an immutable state object for all patient test cases across the app

# Testing Guidance

* Test adding a new patient with the button. Make sure that canceling doesn't save changes, that saving does, etc.
* Test editing an already created patient
* Test deletion
* Try to break stuff

### DIFF
--- a/__tests__/components/CodeEditorModal.test.tsx
+++ b/__tests__/components/CodeEditorModal.test.tsx
@@ -1,0 +1,64 @@
+import '@testing-library/jest-dom';
+import { fireEvent, render, screen, within } from '@testing-library/react';
+import CodeEditorModal, { CodeEditorModalProps } from '../../components/CodeEditorModal';
+import { mantineRecoilWrap } from '../helpers/testHelpers';
+
+describe('CodeEditorModal', () => {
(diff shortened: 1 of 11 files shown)


natjoe4 (changes requested):
All functionality works well and looks good. Just left a few smaller comments

hossenlopp (approved):
Looks like a fantastic start! Only issue is with the editor not warning you when you have invalid JSON. I presume this will come later with validation.

hossenlopp (approved):
Looks great!

natjoe4 (approved):
Looks great to me!

PR #7, opened 2022-06-09
ModelInfo Parsing Script for primaryCodePath 

sarahmcdougall:
# Summary
Creates script that parses the FHIR ModelInfo for the `primaryCodePath` for each resource type, and writes the `primaryCodePath` and other relevant information to a file.

## New behavior
We now have a script that will be used in future tasking to generate the JSON template for the `primaryCodePath` of a given resource type in the TA tool. 

The script writes to the file `parsed-primaryCodePaths.json`. From this file, we can access the `primaryCodePath` for each resource type, as well as its `primaryCodeType`. We can also get a sense of the cardinality of the `primaryCodePath` with the `multipleCardinality` attribute - if true, the cardinality is `0..*` or `1..*`. If false, the cardinality is `0..1` or `1..1`.

Note: there are resource types for which the `primaryCodePath` can take on more than one type. If this is the case, the `primaryCodeType` attribute will be an array whose elements correspond to all the possible types for the `primaryCodePath`. This allows us to later determine a heuristic for which type we want to use. For all these “choice” cases, the cardinality of the possible types is `0..1` or `1..1`, and so the `multipleCardinality` flag is false for all the “choice” cases.

## Code changes
* Bring FHIR ModelInfo into the project
* Create script `parsePrimaryCodePath.ts` that creates output file `fixtures/model-info/parsedPrimaryCodePaths.json`
* Add script and relevant dependencies to `package.json`

# Testing guidance
* Run script with `npm run parse-primaryCodePath`. Ensure that script works and generates the output file.
* Compare the script code to the ModelInfo file to check that we are checking the correct ModelInfo elements.
* Compare the results in the output file to the FHIR documentation. For example, pick a resource type that appears in the output file and check that the `primaryCodeType` is correct for the `primaryCodePath` (check that it matches what appears in the FHIR documentation). Pick a resource that appears in the output file with multiple types, and check that this is the case in the FHIR documentation. Also check the cardinality for these `primaryCodePaths` to make sure the boolean flag accurately reflects the documentation.


p9g:
The idea of the primaryCodePath is to have a default attribute (QDM) or element (FHIR) for a CQL retrieve [terminology filter](http://cql.hl7.org/02-authorsguide.html#filtering-with-terminology). So [MedicationRequest: Foo] could have been written as [MedicationRequest:medication in Foo] where .medication is the primaryCodePath, but I don't know what happens if MedicationRequest.medication is a reference and not a CodeableConcept.

hossenlopp (changes requested):
Nice parsing. Good call on adding the npm script to make it easy to re-run it.

The script file does hit on the lint checks because of the `any` objects from the xml parsing. You could define some types for what those objects have, or just add this to the top of the file.
```
/* eslint-disable @typescript-eslint/no-explicit-any */
```

Additional you may need to export the parsed info as a JS or TS file that can be loaded by normal code. It should possibly live under the `util` directory as well.

hossenlopp (approved):
This looks good! Was able import the primartCodePaths in a `ts-node` REPL with no issues.

PR #8, opened 2022-06-14
Generate Resource Buttons from Data Requirements

natjoe4:
# Summary
Once a measure is uploaded and a test patient is created, a list of test resource buttons is now generated from the data requirements of the measure

## New Behavior
After uploading a measure and creating a test patient, you can now see a list of test resource buttons. The title of the button represents the resource type and the sub title is the name of each valueset included in the codeFilters on the data requirement

## Code Changes
 - Added new `TestResourcesDisplay` component
 - Added function `getDataRequirementFiltersString` to `util/fhir`
 - Updated `measureUploadState` to contain a `valueSetMap` field which maps valueset urls to their names for display purposes
 - Updated `useEffect` of MeasureUpload component to populate a valueSetMap object
 - Added unit testing

# Testing Guidance
 - Run `npm test`
 - Run `npm run dev`
 - Navigate to `http://localhost:3000`
 - Upload a measure
 - Create a test patient
 - Notice the test resources popup and ensure that all data requirements from the uploaded measure are represented
 - Upload a different measure and ensure the list updates


util/fhir.ts
```diff
@@ -36,3 +36,26 @@ export function createPatientResourceString(birthDate: string): string {
...
+ * Identifies the valuesets referenced in a DataRequirement and constructs a string which displays
+ * those valuesets
+ * @param dr {Object} a fhir DataRequirement object
+ * @param valueSetsMap {Object} a mapping of valueset urls to valueset names and titles
+ * @returns {String} displaying the valuesets referenced by a DataRequirement
+ */
+export function getDataRequirementFiltersString(dr: fhir4.DataRequirement, valueSetsMap: any): string {
+  const valueSets = dr.codeFilter?.reduce((acc: string[], e) => {
+    if (e.valueSet) {
+      acc.push(valueSetsMap[e.valueSet]);
```
mgramigna, line 50: I think it might be useful to include the ValueSet OID or full URL in the display, maybe in parentheses? 

```suggestion
      acc.push(`${valueSetsMap[e.valueSet]} (${e.valueSet})`);
```

something like this? Didn't actually test it locally so you don't need to apply the suggestion

__tests__/utils/fhir.test.tsx
```diff
@@ -0,0 +1,74 @@
+import { getDataRequirementFiltersString } from '../../util/fhir';
```
mgramigna, line 1: Thanks for adding tests for this file, I forgot to do it in my PR when the file was initially created. That being said, this should just be a `fhir.test.ts` file, not `tsx`, since it ain't a react component

mgramigna (changes requested):
First round of comments, still fleshing out some thoughts on styling

state/atoms/measureBundle.ts
```diff
@@ -3,6 +3,7 @@ import { atom } from 'recoil';
 interface measureBundleStateType {
   name: string;
   content: fhir4.Bundle | null;
+  valueSetsMap: { [key: string]: string } | null;
```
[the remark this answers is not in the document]
mgramigna: A benefit that this also gives us is that we don't have to explicity compute the valueset map in the `extractMeasureBundle` function, it will just auto update when the state auto updates

__tests__/utils/fhir.test.tsx
natjoe4: good catch

state/atoms/measureBundle.ts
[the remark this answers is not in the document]
natjoe4: Yeah great call. Totally forgot about selectors

components/ResourceCreation/ResourceCreationPanel.tsx
```diff
@@ -0,0 +1,55 @@
...
+  };
+  return (
+    <>
+      <Center>
+        <div style={{ paddingTop: '24px', paddingBottom: '24px' }}>
+          <Button onClick={() => openModal()}>Create Test Patient</Button>
+        </div>
+      </Center>
+      <Grid>
+        {selectedPatient !== null && (
```
mgramigna, line 36: I guess the way we did this, it causes the patient records to be pushed over and resized even if the measure bundle is not uploaded. Maybe we can add a boolean check to both this and the below Grid.Col to make sure that the columns only resize when measureBundle.content is not null?

mgramigna (approved):
yay

PR #9, opened 2022-06-24
Add logic for creating/rending FHIR resources on data element click

sarahmcdougall:
# Summary
Adds logic to create an incomplete FHIR resource when a data element is clicked from the list of rendered data elements. Displays the FHIR resource under the test case for which the FHIR resource was created.

## New Behavior
These changes help satisfy the following user story: “As a user, I want to be able to create test case(s) that will be used as the subjects for measure calculation. It is okay if I have to edit JSON directly as a part of this, but I should be provided a selection of data elements and codes or valuesets to choose from.”

From https://github.com/projecttacoma/fqm-testify/pull/8, the user can select a data element after a test case is created. After clicking on the data element, the TA tool will generate the FHIR resource JSON for the selected data element. The information that appears in the FHIR JSON comes from a combination of the data element's data requirements and the information from the script created in https://github.com/projecttacoma/fqm-testify/pull/7. 

The created resource will first appear in a code editor modal so that the user can make any desired changes. When saved, the new resource will appear under the selected test case. The user has the ability to edit and/or delete the created resource.
## Code Changes
- new `TestResourceCreation` component to handle the creation of new resources and management of the code editor modal for creating resources
- Edits to the `patientTestCase` recoil state to now track an array of resources created for the given test case
- New `selectedDataRequirement` recoil state to track which data element has been clicked from the display - when clicked, this recoil state will trigger the creation of a new resource from the `useEffect()` in the `TestResourceCreation` component
- New function in `util/fhir.ts` to assemble the incomplete FHIR resource. Uses data from the data requirements, and also brings in the measure bundle to grab the ValueSet resource that we want information from. Uses the information from the `parsePrimaryCodePath` script for the given resource type in order to determine which information to bring in and in what format.
- Misc. changes to other `ResourceCreation` components

# Testing Guidance
Run the unit tests, and test out the following scenarios after starting up the app on `localhost:3000`:
- Check that no resources are created until after a measure bundle has been selected and a test case has been made
- Select a data element from the test resource display, and check that the code editor modal appears. Check that the information contained in the JSON is accurate (compare against the FHIR spec for the given resource type)
- Save the new resource, and check that it appears beneath its test case patient
- Try to edit the new resource, and check that the changes are saved
- Try to delete a resource, and check that it actually gets deleted
- Generate multiple test case patients, and multiple resources for each patient. Check that the resources are created for whichever patient is currently selected, and that you can navigate between patients without losing any of the created resources.


__tests__/components/ResourceCreation/TestResourceCreation.test.tsx
```diff
@@ -0,0 +1,204 @@
...
+      mantineRecoilWrap(
+        <>
+          <MockResources />
+          <MockSelectedPatient />
+          <TestResourceCreation />
+        </>
+      )
+    );
+
+    const editResourceButton = screen.getByText(/edit fhir resource/i);
```
hossenlopp, line 105:
```suggestion
    const editResourceButton = screen.getByRole('button', { name: 'Edit FHIR Resource' });
```
Suggest using the role based API for grabbing the button.

__tests__/components/ResourceCreation/TestResourceCreation.test.tsx
```diff
@@ -0,0 +1,204 @@
...
+    fireEvent.click(editResourceButton);
+
+    const modal = screen.getByTestId('code-editor-modal');
```
hossenlopp, line 108:
```suggestion
    const modal = screen.getByRole('dialog', { name: 'Edit FHIR Resource' });
```
Similar to above, but can be done with modals too. This also allows you to not need to add testids to the resources. The testid method for unit testing should only be used as a last resort.

__tests__/components/ResourceCreation/TestResourceCreation.test.tsx
```diff
@@ -0,0 +1,204 @@
...
+      mantineRecoilWrap(
+        <>
+          <MockResources />
+          <MockSelectedPatient />
+          <TestResourceCreation />
+        </>
+      )
+    );
+
+    const resourceInfo = screen.getByText(/test case resources/i);
```
hossenlopp, line 165: Should also check to make sure the Procedure resource is listed.

__tests__/components/ResourceCreation/TestResourceCreation.test.tsx
```diff
@@ -0,0 +1,204 @@
...
+      mantineRecoilWrap(
+        <>
+          <MockResources />
+          <MockSelectedPatient />
+          <TestResourceCreation />
+        </>
+      )
+    );
+
+    const deleteButton = screen.getByText(/delete resource/i) as HTMLButtonElement;
```
hossenlopp, line 196: Similar to above comment. This can use `getByRole`.

hossenlopp (changes requested):
This looks really good. Have a few suggestions for testing though.

hossenlopp (approved):

PR #12, opened 2022-07-06
Fix patient resource editing functionality

hossenlopp:
# Summary

Fixes patient editing so it gives you the resource instead of the recoil state object for the patient. 

## New Behavior

Functional patient editing. No longer wipes out the resources added to the patient when saving a change to the patient.

## Code Changes

Fixes in `PatientCreation.tsx`.

# Testing Guidance

- Create a patient.
- Add resources to the patient.
- Edit patient resource. Note that you just see the Patient resource in the code edit.
- Make a change to the patient (ex. birthday) and save.
- Note that resources are not wiped out.


mgramigna (approved):
Thanks :)

PR #14, opened 2022-07-12
Create Patient References during Resource Creation

hossenlopp:
# Summary

Generate references to the patient when creating resources.

## New Behavior

When a resource is added to the patient's history the appropriate reference to the patient is added to the generated resource.

## Code Changes

Mapping of resourceType to reference attributes was brought over from `deqm-test-server`.
In `util/fhir.ts` the resource creation function now uses the mapping to determine the attribute to put the patient reference on and adds it to the resource.

# Testing Guidance

- Load a measure. EXM130 works well for this because it has `Coverage` types.
- Create a patient.
- Add an Encounter. Check that generated Encounter resource includes `"subject": { "reference": "Patient/{guid}" }`.
- Add a Coverage resource. Check that the generated Coverage resource includes `"beneficiary": { "reference": "Patient/{guid}" }`.


natjoe4 (approved):
Looks good to me and works well!

PR #15, opened 2022-07-13
Add funcionality for importing test patients

mgramigna:
# Summary

This PR allows a user to import test cases into the app via a file upload modal/new button.

## New Behavior

* New button that triggers opening of a modal. User can add a .zip, or 1 to many .json files that are FHIR Bundles to import into the app
  * When uploading a .zip, user can view the contents of that zip via a collapsible panel within the upload modal
* Submitting the import sets the global state such that all imported data appears in the app

## Code Changes

* New component `ImportModal.tsx` that allows the user to upload .zip or .json files. Component will unzip where necessary and ensure that the contents of the file(s) can be passed up to the parent component
* New button in `ResourceCreationPanel.tsx` that renders the above modal and handles submission of the import (including validation of bundles, patient resource existence, etc.)
* Separated out TestCase interfaces in `patientTestCase` atom to make it easier to map Bundles into test cases
* Added helper function for mapping bundles into test cases
* Updated button styling to have create patient be the only blue button. Also changed icons
* Added tests for all the above

# Testing Guidance

**Single/Multiple JSON Upload**

1. Open the app and click "Import Test Patient(s)"
2. Upload 1 to many json files at a time to the dropzone component
3. Hit import -- should see patients show up in the app

Also, worth testing the error cases listed in the ticket:

* Malformed JSON (i.e. an error during JSON.parse)
* Something other than a FHIR Bundle provided (i.e. resourceType !== 'Bundle')
* Bundle is missing a patient resource (i.e. Bundle.entry doesn't contain anything with resourceType 'Patient')

**Zip Archive Upload**

1. Open the app and click "Import Test Patient(s)"
2. Upload a .zip to the upload component (you can obtain a zip by doing "Download All Patients" on some already created patients in the app)
3. You should be able to view the contents of the zip in the collapsible panel if you want
4. Upload should work same as with individual jsons

### DIFF
--- a/components/ResourceCreation/ImportModal.tsx
+++ b/components/ResourceCreation/ImportModal.tsx
@@ -60,11 +60,6 @@ export default function ImportModal({ open, onClose, onImportSubmit }: ImportMod
       opened={open}
       onClose={closeAndReset}
       overflow="outside"
-      styles={{
-        body: {
-          height: '600px'
-        }
-      }}
       title="Import Test Case(s)"
     >
       <Grid>
@@ -88,7 +83,11 @@ export default function ImportModal({ open, onClose, onImportSubmit }: ImportMod
             )}
           </Dropzone>
           <Center>
-            <Group>
+            <Group
+              style={{
+                paddingTop: '12px'
+              }}
+            >
               <Button
                 onClick={() => {
                   onImportSubmit(files);
--- a/components/ResourceCreation/ImportModal.tsx
+++ b/components/ResourceCreation/ImportModal.tsx
@@ -1,6 +1,7 @@
 import { Modal, Button, Center, Group, Grid, Text } from '@mantine/core';
+import { showNotification } from '@mantine/notifications';
 import { Dropzone } from '@mantine/dropzone';
-import { IconFileCheck, IconFileImport } from '@tabler/icons';
+import { IconAlertCircle, IconFileCheck, IconFileImport } from '@tabler/icons';
 import { useState } from 'react';
 import zip from 'jszip';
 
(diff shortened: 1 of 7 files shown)


util/import.ts
```diff
@@ -0,0 +1,22 @@
+import { TestCaseInfo } from '../state/atoms/patientTestCase';
+
+export function bundleToTestCase(bundle: fhir4.Bundle): TestCaseInfo {
+  if (!bundle.entry || bundle.entry.length === 0) {
+    throw new Error('Bundle has no entries');
```
sarahmcdougall, line 5: When we are uploading more than one JSON file at a time, it would be useful to include the name of the file that produced the error. I tested a malformed JSON and it was nice that I was able to see which file produced the parsing error. With the non-Bundle and non-patient resource errors, we don't automatically know which JSON produced the error. This isn't a big deal when we are only uploading a few files, but when we have a lot of files it takes some time to compare the successful file names to the group of files we tried to upload and see which bundle threw the error.

I also noticed that for the other errors in the resource panel (like malformed JSON), if one of the bundles fails, the app will still attempt to import all the remaining bundles. However, for these errors, if the error is thrown, then none of the remaining JSON files will be imported. I tried importing 6 JSON files, where the fourth was missing bundle entries, and I got 3 success and 1 failure rather than 5 success and 1 failure. Should we try and change the logic here to reflect that? Maybe add "Success: __, Failed: __, Unattempted: ___" (or some other wording), or group them into the "Failed?"

sarahmcdougall (changes requested):
Great work on this! The unit tests are very thorough and the UI design is very intuitive. I like the new icons for the buttons :)

Great job handling all the different use cases/error cases here. I have a few suggestions for some of the cases that I tested out, but overall works well!

components/ResourceCreation/ImportModal.tsx
```diff
@@ -0,0 +1,148 @@
...
+        id: 'invalid-zip-upload',
+        icon: <IconAlertCircle />,
+        title: 'Invalid Zip Upload',
+        message: 'Please only upload one or more JSON files or only one zip file',
+        color: 'red'
+      });
+      setShowZipFileExpansion(false);
+    } else {
+      setFiles(uploadedFiles);
+      setFileDisplay(uploadedFiles.map(f => f.name).join(', '));
```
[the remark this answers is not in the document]
mgramigna: Pushed up a change for this yonder: https://github.com/projecttacoma/fqm-testify/pull/15/commits/65be48070cef015ae6203b840bd720198bbd3a21

util/import.ts
mgramigna: Oh shoot yeah. The errors that are thrown here get into the root `catch` block of our `Promise.all`. I talked to @elsaperelli about this, she's gonna add a try catch to make sure these specific errors fail safely as well. They should be grouped into failures

sarahmcdougall (changes requested):

sarahmcdougall (approved):
:clap:

PR #20, opened 2022-07-25
Aligned edit and delete test case resource buttons 

elsaperelli:
## Summary
This PR consolidates FHIR resource buttons to be right-aligned icon buttons. 

## New Behavior
The code behaves the same, the delete resource and edit resource buttons are now just right-aligned and are icons. The resources also alternate colors (white and light gray). The unit tests for these buttons were updated to reflect the icon change. 

## Code Changes
TestResourceCreation.tsx was changed to include the styling changes described above. TestResourcereation.test.tsx was changed to reflect the button changes.

## Testing Guidance
1. Upload a measure bundle and create a test patient.
2. Click on the patient created and add some FHIR resources to the patient by selecting them on the side and clicking save. 
3. Do this for a few FHIR resources- they should be displayed under the test patient in a list that has alternating colors and whose delete and edit buttons are now right-aligned icons.
4. Re-size the window a bit to make sure that the display is relatively consistent and doesn't look horrible.

### DIFF
--- a/__tests__/components/ResourceCreation/TestResourceCreation.test.tsx
+++ b/__tests__/components/ResourceCreation/TestResourceCreation.test.tsx
@@ -104,7 +104,7 @@ describe('TestResourceCreation', () => {
       )
     );
 
-    const editResourceButton = screen.getByRole('button', { name: 'Edit FHIR Resource' });
+    const editResourceButton = screen.getByTestId('edit-resource-button') as HTMLButtonElement;
     fireEvent.click(editResourceButton);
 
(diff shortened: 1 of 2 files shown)


sarahmcdougall (approved):
Love that we are spicing things up with the alternating colors. Tested with different screen sizes and it seems easy to navigate to the edit/delete buttons even with a larger screen. I imagine this will look a little more aesthetically pleasing once we add the value sets to the created resources, getting rid of all the empty space. Unit tests all pass :thumbsup:

PR #22, opened 2022-08-01
Measure calculation functionality

elsaperelli:
## Summary
This PR adds the ability to run measure calculation on a test case for a given bundle and see measure logic highlighting.

## New Behavior
When a user creates a new patient, the user has the option to run the measure calculation by clicking the "Calculate" button that is now shown when a test patient is selected. When Calculate is clicked, the measure calculations are run and the measure logic highlighting is displayed for that patient. There is also a "Hide Results" Button that closes the logic highlighting and becomes a "Show Results" button. 

## Code Changes
- `PatientCreation.tsx`: the calculate function uses fqm-execution function calculateMeasureReports when the calculate button is clicked. 
- `patientTestCase.tsx`: The TestCaseInfo interface now has a third property which is optional for a measureReport.
- `fhir.ts`: The coding of a FHIR.CodeableConcept attribute is changed to be an array.
- `fhir.test.ts`: The unit tests were changed to reflect the change to `fhir.ts`.
- `PatientCreation.test.tsx`: Unit tests added for the changes to `PatientCreation.tsx`.
- `MockMeasureBundle.json`: Added to the fixtures directory for use in the the unit tests. 

## Testing Guidance
1. Upload a measure bundle.
2. Create 2 new test patients. 
3. Select one of the patients. Click "Calculate". The measure logic highlighting should appear as well as a "Hide Results" button.
4. Click the "Hide Results" Button. The results should disappear and that button should now say "Show Results".
5. Add some resources to the test patient from the left panel. 
6. Click "Calculate" again and look for changes in the measure logic highlighting.
7. Click on the other patient you created. There should be a "Calculate" button, but no "Hide Results" or "Show Results" button.

### DIFF
--- a/__tests__/components/ResourceCreation/PatientCreation.test.tsx
+++ b/__tests__/components/ResourceCreation/PatientCreation.test.tsx
@@ -1,4 +1,4 @@
-import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { act, fireEvent, render, screen, waitFor } from '@testing-library/react';
 import '@testing-library/jest-dom';
 import { mantineRecoilWrap, getMockRecoilState } from '../../helpers/testHelpers';
 import PatientCreation from '../../../components/ResourceCreation/PatientCreation';
@@ -6,14 +6,18 @@ import { patientTestCaseState } from '../../../state/atoms/patientTestCase';
 import { download } from '../../../util/downloadUtil';
 import { selectedPatientState } from '../../../state/atoms/selectedPatient';
 import { measureBundleState } from '../../../state/atoms/measureBundle';
-import testBundle from '../../fixtures/bundles/MockMeasureBundle.json';
 import MeasureUpload from '../../../components/MeasureUpload';
 import { Calculator } from 'fqm-execution';
-import { act } from 'react-dom/test-utils';
 
 jest.mock('../../../util/downloadUtil', () => ({
   download: jest.fn()
 }));
+
+const MOCK_MEASURE_BUNDLE: fhir4.Bundle = {
+  resourceType: 'Bundle',
+  type: 'collection',
+  entry: []
+};
 
 const MOCK_MEASURE_REPORT: fhir4.MeasureReport = {
   resourceType: 'MeasureReport',
@@ -199,7 +203,7 @@ describe('PatientCreation', () => {
 
     const MockMB = getMockRecoilState(measureBundleState, {
       name: 'testName',
-      content: testBundle as fhir4.Bundle
+      content: MOCK_MEASURE_BUNDLE
     });
 
     // Mock calculate data requirements because of the changing state
@@ -215,32 +219,34 @@ describe('PatientCreation', () => {
       results: [MOCK_MEASURE_REPORT]
     });
 
-    await act(async () => {
-      render(
-        mantineRecoilWrap(
-          <>
-            <MockPatients />
-            <MockSelectedPatient />
-            <MockMB />
-            <MeasureUpload />
-            <PatientCreation {...DEFAULT_PROPS} isPatientModalOpen={false} currentPatient={null} />
-          </>
-        )
-      );
-    });
+    render(
+      mantineRecoilWrap(
+        <>
+          <MockPatients />
+          <MockSelectedPatient />
+          <MockMB />
+          <MeasureUpload />
+          <PatientCreation {...DEFAULT_PROPS} isPatientModalOpen={false} currentPatient={null} />
+        </>
+      )
+    );
 
-    const calculateButton = screen.getByTestId('calculate-button') as HTMLButtonElement;
+    const calculateButton = screen.getByRole('button', { name: 'Calculate' }) as HTMLButtonElement;
     expect(calculateButton).toBeInTheDocument();
 
-    const toggleShowCalculationButton = screen.queryByTestId('toggle-show-calculation-button') as HTMLButtonElement;
+    const toggleShowCalculationButton = screen.queryByRole('button', {
+      name: 'Show Logic Highlighting'
+    }) as HTMLButtonElement;
     expect(toggleShowCalculationButton).not.toBeInTheDocument();
 
     await act(async () => {
       fireEvent.click(calculateButton);
     });
 
-    await waitFor(() => {
-      const toggleShowCalculationButton = screen.getByTestId('toggle-show-calculation-button') as HTMLButtonElement;
+    waitFor(() => {
+      const toggleShowCalculationButton = screen.getByRole('button', {
+        name: 'Hide Logic Highlighting'
+      }) as HTMLButtonElement;
       expect(toggleShowCalculationButton).toBeInTheDocument();
       const textDiv = screen.getByText('test123');
       expect(textDiv).toBeInTheDocument();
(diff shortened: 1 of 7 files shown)


__tests__/components/ResourceCreation/PatientCreation.test.tsx
```diff
@@ -4,11 +4,32 @@ import { mantineRecoilWrap, getMockRecoilState } from '../../helpers/testHelpers
 import PatientCreation from '../../../components/ResourceCreation/PatientCreation';
 import { patientTestCaseState } from '../../../state/atoms/patientTestCase';
 import { download } from '../../../util/downloadUtil';
+import { selectedPatientState } from '../../../state/atoms/selectedPatient';
+import { measureBundleState } from '../../../state/atoms/measureBundle';
+import testBundle from '../../fixtures/bundles/MockMeasureBundle.json';
+import MeasureUpload from '../../../components/MeasureUpload';
+import { Calculator } from 'fqm-execution';
+import { act } from 'react-dom/test-utils';
```
mgramigna, line 12: I _think_ it's the same thing (not sure), but we usually import `act` from `@testing-library/react` for consistency. Can you just add act to the existing import on line 1 and remove this statement? It should work the same

util/fhir.ts
```diff
@@ -160,7 +160,8 @@ export function createFHIRResourceString(
   let codeData;
   if (primaryCodeType === 'FHIR.CodeableConcept') {
     codeData = {
-      coding: coding
+      // Need to add coding as an array for codeable concept
+      coding: [coding]
```
mgramigna, line 164: Nitpick but can you make the type of `codeData` above `fhir4.CodeableConcept | fhir4.Coding | string | undefined | null`? That should prevent these kinds of errors in this function in the future, and it should still work

components/ResourceCreation/PatientCreation.tsx
```diff
@@ -176,8 +230,32 @@ function PatientCreation({
...
+                    <Button data-testid="calculate-button" onClick={() => clickCalculateButton(id)} color="gray">
+                      Calculate
+                    </Button>
+                    {currentPatients[id].measureReport !== undefined && (
+                      <Button
+                        data-testid="toggle-show-calculation-button"
+                        onClick={() => setShowCalculation(o => !o)}
+                        color="gray"
+                      >
+                        {showCalculation === true ? `Hide Results` : `Show Results`}
```
mgramigna, line 242: I think a button name like "Show Logic Highlighting"/"Hide Logic Highlighting" or something makes more sense so the user knows what to expect?

__tests__/components/ResourceCreation/PatientCreation.test.tsx
```diff
@@ -152,4 +176,74 @@ describe('PatientCreation', () => {
...
+            <MockSelectedPatient />
+            <MockMB />
+            <MeasureUpload />
+            <PatientCreation {...DEFAULT_PROPS} isPatientModalOpen={false} currentPatient={null} />
+          </>
+        )
+      );
+    });
+
+    const calculateButton = screen.getByTestId('calculate-button') as HTMLButtonElement;
```
mgramigna, line 232: In general we shouldn't default to using a `testid` where possible. One of the main paradigms of react unit testing is to test things the user sees (for example, a button that says "Calculate"). A user can't see a testid, and so it's use should really be for when we don't have another way to access this element within a unit test. Try this instead, does it still work?

``` typescript
const calculateButton = screen.getByRole('button', { name: 'Calculate' }) as HTMLButtonElement
```

does it still work?

__tests__/components/ResourceCreation/PatientCreation.test.tsx
```diff
@@ -152,4 +176,74 @@ describe('PatientCreation', () => {
...
+    expect(calculateButton).toBeInTheDocument();
+
+    const toggleShowCalculationButton = screen.queryByTestId('toggle-show-calculation-button') as HTMLButtonElement;
```
mgramigna, line 235: same comment about the Role being preferred over testid

__tests__/components/ResourceCreation/PatientCreation.test.tsx
```diff
@@ -152,4 +176,74 @@ describe('PatientCreation', () => {
...
+    expect(toggleShowCalculationButton).not.toBeInTheDocument();
+
+    await act(async () => {
+      fireEvent.click(calculateButton);
+    });
+
+    await waitFor(() => {
+      const toggleShowCalculationButton = screen.getByTestId('toggle-show-calculation-button') as HTMLButtonElement;
```
mgramigna, line 243: same comment about the Role being preferred over testid

__tests__/fixtures/bundles/MockMeasureBundle.json
```diff
@@ -0,0 +1,5 @@
+{
```
mgramigna, line 1: nitpick, but I don't think this warrants its own file since it's so small. You can probably just make this a `const MOCK_MEASURE_BUNDLE` at the top of the test file where you use it

components/ResourceCreation/PatientCreation.tsx
```diff
@@ -34,6 +45,46 @@ function PatientCreation({
...
+    } catch (error) {
+      console.log(error);
+    }
+  };
+
+  // Function to calculate the selected patient's measure report
+  const calculate = async (id: string | null) => {
+    const options: CalculatorTypes.CalculationOptions = {
+      calculateHTML: true,
+      calculateSDEs: false,
```
mgramigna, line 65: I remember we set this to false because of a bug in fqm-execution. This was actually fixed (not released yet but hopefully will be soon), can we just add a `TODO` comment above this to flip `calculateSDEs` to `true` once a new `fqm-execution` version is released/the dependency is updated? I don't wanna forget

components/ResourceCreation/PatientCreation.tsx
```diff
@@ -34,6 +45,46 @@ function PatientCreation({
...
+  const measureBundle = useRecoilValue(measureBundleState);
+  const [showCalculation, setShowCalculation] = useState(false);
+  const { classes } = useStyles();
+
+  // Function to wrap the calculate function and catch errors in fqm-execution
+  const clickCalculateButton = async (id: string | null) => {
+    try {
+      await calculate(id);
+    } catch (error) {
+      console.log(error);
```
mgramigna, line 57: Can we do something better than just silently console log an error here? Ideally the user can become aware if there are any issues with calculation for a patient.

We already have precedent in this codebase for showing notifications/have the infrastructure set up for it. Can we make use of `showNotification` here instead to have the error message pop up on the screen? See the ImportModal component for examples of where this is done

components/ResourceCreation/PatientCreation.tsx
```diff
@@ -34,6 +45,46 @@ function PatientCreation({
...
+      reportType: 'individual',
+      measurementPeriodStart: measurementPeriod.start?.toISOString(),
+      measurementPeriodEnd: measurementPeriod.end?.toISOString()
+    };
+
+    if (id && measureBundle.content) {
+      const patientBundle = createPatientBundle(
+        currentPatients[id].patient,
+        currentPatients[id].resources
+      ) as fhir4.Bundle;
```
mgramigna, line 75:
```suggestion
      );
```

The `createPatientBundle` function already returns a `fhir4.Bundle`, so this case is unnecessary. We should only use type casting when it is impossible for typescript to determine the type of something (this is why we do it in the unit tests sometimes. TypeScript doesn't know anything about the "type" of an arbitrary JSON file, so we use a cast because we know it's a bundle)

mgramigna (changes requested):
This is great! Functionality looks good and is well tested. I have a bunch of small inline comments, but overall great work :)

mgramigna (approved):

PR #23, opened 2022-08-02
Measure Calculation on All Patients

sarahmcdougall:
# Summary
Adds functionality for the user to run measure calculation on all test cases and see the population results in a tabular format.

## New Behavior
This PR corresponds to the following user story: “As a user, I want to be able to run measure calculation on my test case for a given measure and see a human-readable table that shows the population results for every patient I used during the calculation”

Adds a button (“Calculate Population Results”) that allows the user to run measure calculation of all the created patients for the uploaded FHIR Measure. The population results for each patient appear in a population results table, with a row for each patient, labeled with the patient’s name and date of birth. This table appears as a popover so that the user can still view the current patients and their resources.

## Code Changes
* New component - `PopulationCalculation` - with functions for gathering all the patient names/DOBs, calculating measure reports for each patient using fqm-execution, throwing errors from fqm-execution in a Mantine notification, etc.
    * “Calculate Population Results” button appears when there is at least one patient, and a measure bundle has been uploaded
    * Population Results popover appears when the calculation ends and is successful
* Updated `PatientCreation` component to be scrollable on overflow (allows for more room for displaying population results table and calculation button)
* Bring in `ecqm-visualizers` as a dependency

# Testing Guidance
Run unit tests with `npm run test`. Then, try the following in the app:
* Start the app and check that the “Calculate Population Results” button is not present
* Add at least one patient and upload a measure bundle, and then check that the “Calculate Population Results” button appears at the bottom of the page, below the resource creation panel
* Try creating patients that will fall into the numerator/denominator/IPP populations (I found it easiest to test this by uploading the EXM130 measure bundle, then uploading the EXM130 numerator/denominator patient bundles and making sure the measurement period is for 1/1/2019-12/31/2019)
* After adding some patients, click the “Calculate Population Results” button and check that the table renders correctly in the popover
* Check that the popover can be closed when clicking on the “Close” button

### Other Notes
The way in which we display the population results table can be changed. We want to be able to still view the current patients while having the table on the screen, so we don’t want to route to a new page, and displaying the table at the bottom of the screen may make it hard for the user to reference the patients while looking at the results table. I’m not super happy with the popover, so I am open to suggestions on how best to display the table, especially when there is a large number of patients that are performing calculation on.


components/PopulationCalculation.tsx
```diff
@@ -0,0 +1,137 @@
...
+                          size="sm"
+                          color="red"
+                        >
+                          Close
+                        </Button>
+                      </Group>
+                      <div data-testid="results-table">
+                        <PopulationResultsViewer reports={measureReports} />
+                      </div>
+                    </Popover>
```
elsaperelli, line 127: I think this is a great design to start but I think we can make it even better even though this is a difficult UI challenge since there are so many things we want the user to be able to see at the same time. My first thought was to put the population result table to the right of the patient test cases so that under the buttons you see:
Resources | Patients | Population Results. I am not _quite_ sure how exactly this would look but I will make some sketches and share if that would be helpful! This is also where a second opinion from @mgramigna or @hossenlopp would be super helpful!

elsaperelli (changes requested):
This is great! I see why the UI is so challenging here. I just have one code suggestion and one design suggestion.

__tests__/components/PopulationCalculation.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+  it('should not render Calculate Population Results button by default', () => {
+    render(
+      mantineRecoilWrap(
+        <>
+          <PopulationCalculation />
+        </>
+      )
+    );
+
+    const calculateButton = screen.queryByTestId('calculate-all-button');
```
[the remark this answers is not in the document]
sarahmcdougall: When I try `queryByRole` when I'm testing for something that is not in the document, I get the following: 

"TestingLibraryElementError: Unable to find an accessible element with the role "button" and name "Calculate Population Results" There are no accessible roles. But there might be some inaccessible roles. If you wish to access them, then set the `hidden` option to `true`."

I don't have a lot of experience with frontend testing, so I used other `not.toBeInTheDocument()` unit tests as precedent, and they all seem to use `queryByTestId()`. Maybe it worked for your unit test because you did have a button present, but not with the specific name (since yours changes from "show" to "hide")? Not really sure

components/PopulationCalculation.tsx
```diff
@@ -0,0 +1,137 @@
...
+   * the population results. Catches errors in fqm-execution that result from calculateMeasureReports().
+   */
+  const runCalculation = () => {
+    calculateMeasureReports()
+      .then(measureReports => {
+        if (measureReports) {
+          const patientLabels = createPatientLabels();
+          const labeledMeasureReports: DetailedMeasureReport[] = [];
+          measureReports.forEach((mr, i) => {
+            labeledMeasureReports.push({ label: patientLabels[i], report: mr as fhirJson.MeasureReport });
```
natjoe4, line 74: Can we always rely on the measureReports for each patient being returned in the same order that the patients are in the passed-in bundle? It seems like that is what this logic is relying on

natjoe4 (changes requested):
Works great! I have some thoughts on the popover. The file upload takes up waaaaay more space than I think it needs to, and seems like a great size and shape to display population results. I'm wondering if we should change the measureBundle upload to a button and put it on the line with the patient creation button? Then save that top space for Population results?? What do you think?

components/PopulationCalculation.tsx
mgramigna: prob shouldn't. Should look at the mr subject to ensure that patient ID matches

sarahmcdougall:
> Works great! I have some thoughts on the popover. The file upload takes up waaaaay more space than I think it needs to, and seems like a great size and shape to display population results. I'm wondering if we should change the measureBundle upload to a button and put it on the line with the patient creation button? Then save that top space for Population results?? What do you think?

Ooooh good idea. I agree that the measure upload does take up more space than it used to. I'm hesitant to put the population results at the very top of the page because when I think of the flow of the app I think of putting measure upload/measurement period at the top, then all the patient import/export functionality, then all the resource creation functionality, and then finally the population calculation at the bottom so that the user can click on that once they are happy with everything else. I guess this flow won't always be the case though, like if a user looks at the population results and then wants to tweak some test cases - then it might be useful to have it at the top. I can try playing around with it and see how it looks :)

natjoe4 (approved):
Great work. I am team drawer

elsaperelli (approved):
Looks great to me!

PR #25, opened 2022-08-09
Static patient name lookup

mgramigna:
# Summary

Moves the generation of fake patient names to a dev script rather than during realtime. See this snippet from [faker usage docs](https://fakerjs.dev/guide/#browser):

> Using the browser is great for experimenting 👍. However, due to all of the strings Faker uses to generate fake data, Faker is a large package. It's > 5 MiB minified. Please avoid deploying Faker in your web app.

This PR listens to them and generates the fake patient names at build time rather than at runtime.

## New Behavior

None in the app. New script you can run now `npm run build:names`

## Code Changes

* Added `scripts/generatePatientNames.ts` to generate a list of patient names by gender
* Add data/names.ts for lookups
* Update util/randomizer.ts and util/fhir.ts to not use faker library and just pull from existing lookup
* Move faker to dev dependencies
* Add npm script to build patient names
* Update build script to run the relevant files in `scripts` before doing a production build
* Add `scripts/generatePatientNames.ts` to be excluded in typescript compilation so faker doesn't ship with the app

# Testing Guidance

Test both `npm run build:names` and `npm run build`, you should see the output.

Test creating patients in the app and make sure the names/genders are still randomized and appropriate


package.json
```diff
@@ -5,7 +5,8 @@
   "scripts": {
     "check": "npm run test && npm run lint && npm run prettier",
     "dev": "next dev -p 3000",
-    "build": "next build && npm run parse-primaryCodePath",
+    "build": "npm run parse-primaryCodePath && npm run build:names && next build",
```
lmd59, line 8: Should we actually run build:names as part of a basic build? Seems like since the names file is checked in, we probably want to just leave as is except in exceptional circumstances.

mgramigna: Yeah that sounds right actually. I wonder if I should remove both that and the primary code path one since both results of those scripts are already checked into the repo?

mgramigna: @sarahmcdougall Do we def need the primary code path parsing to happen in the build script or is it just there for convenience?

sarahmcdougall: @mgramigna nah we just decided to put it in there for convenience. It doesn't need to happen in the build script

data/names.ts
```diff
@@ -0,0 +1,160 @@
...
+      'Antoinette825',
+      'Veronica857',
+      'Jo479',
+      'Tasha428',
+      'Lindsay585',
+      'Daisy998',
+      'Holly148',
+      'Claire251',
+      'Barbara422',
+      'Sarah450',
```
sarahmcdougall, line 80: nice

tsconfig.json
```diff
@@ -13,8 +13,8 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "jsx": "preserve",
-    "incremental": true,
+    "incremental": true
   },
   "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
-  "exclude": ["node_modules"]
+  "exclude": ["node_modules", "scripts/generatePatientNames.ts"]
```
sarahmcdougall, line 19: Why do we need to exclude the script from tsconfig?

sarahmcdougall (commented):
This works really well and makes npm install soooo much faster! No suggested changes but I do agree with Lauren that maybe we should consider removing the script (and the parse primary code path script) from the base build.

tsconfig.json
mgramigna: Good question. I am excluding this because the scripts should not be included with the production build of the application.

When we run `next build`, it's going to take everything in `include` and use that as the typescript source files to generate the built application. The `scripts` directory should be explicitly excluded from this since it only used during development and not for the actual functionality in the application.

I was going to exclude the whole directory, but I realized that we're actually using a TypeScript interface exported from the parsePrimaryCodePath script in the source code, so excluding it will break the type defs. We should resolve this in a separate task

mgramigna: Although I'm happy to tackle it here just lmk

sarahmcdougall: Thanks for the explanation :) I was curious as to why we would exclude the one script but not the other, so that explanation makes sense. You can address it here if you want but I am also fine with it being a separate task 👍

mgramigna:
@sarahmcdougall I pushed up a fix for the above thread, I'm gonna amend my testing instructions also to say:

```
npm run build
npm start
```

This will run the production build of the app, which we should make sure works with the recent tsconfig changes :)

sarahmcdougall (approved):
Ran the production build and that works well, and I ran all the scripts and they produce the correct output. I like the script consistency that we have now. ~ L-G-T-M ~

lmd59 (approved):
LGTM2!

PR #26, opened 2022-08-10
Resource code and display name added

elsaperelli:
## Summary
Display the code and the first few words of the display text next to the test case resources in order to differentiate between resources that are of the same type. 

## New Behavior
When the display text is too long for the page, the text is cut off by an ellipses. To see the full text, the use can hover over the text with their mouse and the full text will appear in a popup "tooltip". 

## Code Changes
- `fhir.ts`: function added that takes a `Fhir4Resource` as a parameter and returns a string of that resource's code and display text. 
- `TestResourceCreation`: the `getFhirResourceString` in `fhir.ts` is called in order to display the code and display text next to the resourceType. In addition, this text is wrapped in a tooltip component so that the user can hover over the text to see the full text if it is cut off.
- `TestResourceCreation.test.tsx`: unit tests were added to test these changes.
- fhirpath added as a dependency.

## Testing Guidance
Run unit tests with `npm run test`. Then, try the following in the app:
1. Upload a measure bundle JSON file. 
2. Create a test patient.
3. Add some test case resources to that patient. The test case resources should display "ResourceType (code: display)".
4. Click the Edit button. Make sure that the values in the JSON for code and display match the ones shown. 
5. Add test case resources until you get one where the text is so long that it is cut off on the page with ellipses. 
6. Hover over the cut off text. You should see a popup that contains the full display text of the test case resource.

### DIFF
--- a/__tests__/components/ResourceCreation/TestResourceCreation.test.tsx
+++ b/__tests__/components/ResourceCreation/TestResourceCreation.test.tsx
@@ -1,4 +1,4 @@
-import { act, fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { fireEvent, render, screen } from '@testing-library/react';
 import '@testing-library/jest-dom';
 import { mantineRecoilWrap, getMockRecoilState } from '../../helpers/testHelpers';
 import TestResourceCreation from '../../../components/ResourceCreation/TestResourceCreation';
@@ -137,7 +137,7 @@ describe('TestResourceCreation', () => {
     expect(resourceInfo).not.toBeInTheDocument();
   });
 
-  it('should render test resource list when populated', async () => {
+  it('should render test resource list when populated', () => {
     const MockResources = getMockRecoilState(patientTestCaseState, {
       'example-test-case': {
         patient: {
@@ -164,7 +164,7 @@ describe('TestResourceCreation', () => {
     });
     const MockSelectedPatient = getMockRecoilState(selectedPatientState, 'example-test-case');
 
-    const baseDom = render(
+    render(
       mantineRecoilWrap(
         <>
           <MockResources />
--- a/__tests__/components/ResourceCreation/TestResourceCreation.test.tsx
+++ b/__tests__/components/ResourceCreation/TestResourceCreation.test.tsx
@@ -164,7 +164,7 @@ describe('TestResourceCreation', () => {
     });
     const MockSelectedPatient = getMockRecoilState(selectedPatientState, 'example-test-case');
 
-    render(
+    const baseDom = render(
       mantineRecoilWrap(
         <>
           <MockResources />
@@ -182,15 +182,6 @@ describe('TestResourceCreation', () => {
 
     const resourceCode = screen.getByText(/123: Colectomy/i);
     expect(resourceCode).toBeInTheDocument();
-
-    await act(async () => {
-      fireEvent.mouseOver(resourceCode);
-    });
-
-    waitFor(() => {
-      const resourceCodes = screen.getAllByText(/123: Colectomy/i);
-      expect(resourceCodes).toHaveLength(2);
-    });
 
     const editResourceButton = screen.getByTestId('edit-resource-button') as HTMLButtonElement;
     expect(editResourceButton).toBeInTheDocument();
--- a/__tests__/components/ResourceCreation/TestResourceCreation.test.tsx
+++ b/__tests__/components/ResourceCreation/TestResourceCreation.test.tsx
@@ -1,4 +1,4 @@
-import { fireEvent, render, screen } from '@testing-library/react';
+import { act, fireEvent, render, screen, waitFor } from '@testing-library/react';
 import '@testing-library/jest-dom';
 import { mantineRecoilWrap, getMockRecoilState } from '../../helpers/testHelpers';
 import TestResourceCreation from '../../../components/ResourceCreation/TestResourceCreation';
@@ -137,7 +137,7 @@ describe('TestResourceCreation', () => {
     expect(resourceInfo).not.toBeInTheDocument();
   });
 
-  it('should render test resource list when populated', () => {
+  it('should render test resource list when populated', async () => {
     const MockResources = getMockRecoilState(patientTestCaseState, {
       'example-test-case': {
         patient: {
@@ -149,7 +149,15 @@ describe('TestResourceCreation', () => {
             resourceType: 'Procedure',
             id: 'test-id',
             status: 'completed',
-            subject: {}
+            subject: {},
+            code: {
+              coding: [
+                {
+                  code: '123',
+                  display: 'Colectomy'
+                }
+              ]
+            }
           }
         ]
       }
@@ -171,6 +179,18 @@ describe('TestResourceCreation', () => {
 
     const procedureResource = screen.getByText(/1. Procedure/i);
     expect(procedureResource).toBeInTheDocument();
+
+    const resourceCode = screen.getByText(/123: Colectomy/i);
+    expect(resourceCode).toBeInTheDocument();
+
+    await act(async () => {
+      fireEvent.mouseOver(resourceCode);
+    });
+
+    waitFor(() => {
+      const resourceCodes = screen.getAllByText(/123: Colectomy/i);
+      expect(resourceCodes).toHaveLength(2);
+    });
 
     const editResourceButton = screen.getByTestId('edit-resource-button') as HTMLButtonElement;
     expect(editResourceButton).toBeInTheDocument();
--- a/__tests__/utils/fhir.test.ts
+++ b/__tests__/utils/fhir.test.ts
@@ -164,10 +164,10 @@ const TEST_MEASURE_BUNDLE_WITH_EXPANSION: fhir4.Bundle = {
 };
 
 const RESOURCE_WITH_NO_SUMMARY: fhir4.Resource = {
-  resourceType: 'Prodcedure'
+  resourceType: 'Procedure'
 };
 
-const RESOURCE_WITH_NO_CODE_PATH: fhir4.Resource = {
+const RESOURCE_WITH_NO_CODE: fhir4.Resource = {
   resourceType: 'Procedure',
   id: 'procedure-id'
 };
@@ -242,6 +242,22 @@ const PROCEDURE_RESOURCE_WITH_TWO_CODES: fhir4.Procedure = {
   }
 };
 
+const MEASURE_REPORT_WITH_ID: fhir4.MeasureReport = {
+  resourceType: 'MeasureReport',
+  id: 'measure-report-id',
+  type: 'individual',
+  status: 'complete',
+  measure: '',
+  period: {
+    start: '',
+    end: ''
+  },
+  text: {
+    div: 'test123',
+    status: 'additional'
+  }
+};
+
 describe('getDataRequirementFiltersString', () => {
   test('returns an empty string for resource with no valuesets', () => {
     expect(getDataRequirementFiltersString(DATA_REQUIREMENT_WITH_NO_VALUE_SETS, VS_MAP)).toEqual('');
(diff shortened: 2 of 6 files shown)


util/fhir.ts
```diff
@@ -37,6 +39,14 @@ export function createPatientResourceString(birthDate: string): string {
   return JSON.stringify(pt, null, 2);
 }
 
+export function getFhirResourceString(resource: fhir4.Resource) {
+  const primaryCodePath = parsedPrimaryCodePaths[resource.resourceType].primaryCodePath;
+  return `(${fhirpath.evaluate(resource, `${primaryCodePath}.coding.code`)}: ${fhirpath.evaluate(
```
hossenlopp, line 44: If the code at the path is not populated, this just returns `(: )`. Suggest returning perhaps the `id` of the resource or nothing.

util/fhir.ts
```diff
@@ -37,6 +39,14 @@ export function createPatientResourceString(birthDate: string): string {
   return JSON.stringify(pt, null, 2);
 }
 
+export function getFhirResourceString(resource: fhir4.Resource) {
+  const primaryCodePath = parsedPrimaryCodePaths[resource.resourceType].primaryCodePath;
```
hossenlopp, line 43: If the resource doesn't have a primaryCodePath then this throws an error. Check to make sure it has one and fallback to returning something else. Also discussed below.

util/fhir.ts
```diff
@@ -37,6 +39,14 @@ export function createPatientResourceString(birthDate: string): string {
   return JSON.stringify(pt, null, 2);
 }
 
+export function getFhirResourceString(resource: fhir4.Resource) {
```
hossenlopp, line 42: The name of this function seems to imply it returns the string for the entire resource. This kind of returns something like a summary or title for the resource. Also since this is a reusable function, add JSDoc comments to explain it.

__tests__/components/ResourceCreation/TestResourceCreation.test.tsx
```diff
@@ -172,6 +180,18 @@ describe('TestResourceCreation', () => {
...
 
+    const resourceCode = screen.getByText(/123: Colectomy/i);
+    expect(resourceCode).toBeInTheDocument();
+
+    await act(async () => {
+      fireEvent.mouseOver(resourceCode);
+    });
+
+    waitFor(() => {
+      const resourceCodes = screen.getAllByText(/123: Colectomy/i);
```
hossenlopp, line 191: Suggest using the `getByRole` function instead of `getAllByText` to make sure that something that is of the `tooltip` ARIA role comes up. This makes the testing act more like it is a user using the page.

hossenlopp (changes requested):
Looks good but can fail when data it looks for doesn't exist.

util/fhir.ts
elsaperelli: Does `getFhirResourceSummary` sound better?

util/fhir.ts
```diff
@@ -6,6 +6,8 @@ import { parsedPrimaryCodePaths } from './primaryCodePaths';
 import _ from 'lodash';
 import { ReferencesMap } from './referencesMap';
 
+const fhirpath = require('fhirpath');
```
sarahmcdougall, line 9: I think we can change this to `import fhirpath from 'fhirpath';` for consistency

util/fhir.ts
sarahmcdougall: I agree with Hoss's comment. We should rename this function to something that indicates that this returns a title/summary/code and display string.

sarahmcdougall (changes requested):
I agree with all of Hoss's suggested changes 👍 
I'm not sure if this is something that can easily be fixed here, but I noted that the resources with an ellipsis seem to be a little bigger than the ones without an ellipsis (in the screenshot below it's resource 5 compared to resource 6). We might want to look into this for consistency
Screen Shot 2022-08-11 at 10 32 17 AM

util/fhir.ts
```diff
@@ -39,12 +38,26 @@ export function createPatientResourceString(birthDate: string): string {
...
-export function getFhirResourceString(resource: fhir4.Resource) {
-  const primaryCodePath = parsedPrimaryCodePaths[resource.resourceType].primaryCodePath;
-  return `(${fhirpath.evaluate(resource, `${primaryCodePath}.coding.code`)}: ${fhirpath.evaluate(
-    resource,
-    `${primaryCodePath}.coding.display`
-  )})`;
+/**
+ * Identifies the primary code path of a resource and constructs a string which displays
+ * resource summary information depending on what is a available
+ * @param resource {Object} a fhir DataRequirement object
```
sarahmcdougall, line 44: I think we might want to edit this doc string to say "a fhir Resource object" - I believe we are passing in a created resource rather than the data elements

__tests__/components/ResourceCreation/TestResourceCreation.test.tsx
```diff
@@ -1,4 +1,4 @@
-import { fireEvent, render, screen } from '@testing-library/react';
+import { act, fireEvent, render, screen, waitFor } from '@testing-library/react';
```
sarahmcdougall, line 1:
```suggestion
import { fireEvent, render, screen } from '@testing-library/react';
```
We can get rid of these imports since they are no longer being used in this file

__tests__/components/ResourceCreation/TestResourceCreation.test.tsx
```diff
@@ -137,7 +137,7 @@ describe('TestResourceCreation', () => {
     expect(resourceInfo).not.toBeInTheDocument();
   });
 
-  it('should render test resource list when populated', () => {
+  it('should render test resource list when populated', async () => {
```
sarahmcdougall, line 140: I think this doesn't need to be async anymore?

__tests__/components/ResourceCreation/TestResourceCreation.test.tsx
```diff
@@ -149,14 +149,22 @@ describe('TestResourceCreation', () => {
...
+              ]
+            }
           }
         ]
       }
     });
     const MockSelectedPatient = getMockRecoilState(selectedPatientState, 'example-test-case');
 
-    render(
+    const baseDom = render(
```
sarahmcdougall, line 167: Do we need to set this as a const? I am getting an eslint warning since `baseDom` doesn't get called anywhere.

sarahmcdougall (changes requested):

util/fhir.ts
```diff
@@ -34,6 +35,36 @@ export function createPatientResourceString(birthDate: string): string {
...
+ * @param resource {Object} a fhir Resource object
+ * @returns {String} displaying the code and display text, code, or id of the resource or nothing
+ */
+export function getFhirResourceSummary(resource: fhir4.Resource) {
+  const primaryCodePath = parsedPrimaryCodePaths[resource.resourceType]?.primaryCodePath;
+  const resourceCode = `${fhirpath.evaluate(resource, `${primaryCodePath}.coding.code`)}`;
+  const resourceDisplay = `${fhirpath.evaluate(resource, `${primaryCodePath}.coding.display`)}`;
+
+  if (primaryCodePath) {
+    if (resourceCode !== '' && resourceDisplay !== '') {
```
hossenlopp, line 50:
```suggestion
  if (primaryCodePath) {
      const resourceCode = `${fhirpath.evaluate(resource, `${primaryCodePath}.coding.code`)}`;
      const resourceDisplay = `${fhirpath.evaluate(resource, `${primaryCodePath}.coding.display`)}`;
    if (resourceCode !== '' && resourceDisplay !== '') {
```

Suggest moving the code and display grabbing to inside this if so it is only executed if we do have a `primaryCodePath`.

util/fhir.ts
```diff
@@ -34,6 +35,36 @@ export function createPatientResourceString(birthDate: string): string {
...
+      return `(${fhirpath.evaluate(resource, 'id')})`;
+    } else {
+      return '';
+    }
+  } else if (fhirpath.evaluate(resource, 'id')) {
+    return `(${fhirpath.evaluate(resource, 'id')})`;
+  } else {
+    return '';
+  }
+}
```
hossenlopp, line 66:
```suggestion
    }
  }
  
  if (resource.id)) {
    return `(${resource.id})`;
  } else {
    return '';
  }
}
```

To not repeat the grabbing of the id and making that string, let the situation where we cant find a code fall out of the top level if block and then get picked up by the code at the end. Also fhirpath isn't necessary for grabbing the id.

util/fhir.ts
```diff
@@ -34,6 +35,36 @@ export function createPatientResourceString(birthDate: string): string {
   return JSON.stringify(pt, null, 2);
 }
 
+/**
+ * Identifies the primary code path of a resource and constructs a string which displays
+ * resource summary information depending on what is a available
+ * @param resource {Object} a fhir Resource object
+ * @returns {String} displaying the code and display text, code, or id of the resource or nothing
+ */
+export function getFhirResourceSummary(resource: fhir4.Resource) {
```
hossenlopp, line 44: This function should have some unit tests that test it directly instead of through the component. The tests should be much easier to write to test all the paths through this function.

util/fhir.ts
```diff
@@ -34,6 +35,36 @@ export function createPatientResourceString(birthDate: string): string {
   return JSON.stringify(pt, null, 2);
 }
 
+/**
+ * Identifies the primary code path of a resource and constructs a string which displays
+ * resource summary information depending on what is a available
+ * @param resource {Object} a fhir Resource object
```
hossenlopp, line 41:
```suggestion
 * @param resource {fhir4.Resource} a fhir Resource object
```

hossenlopp (changes requested):
More suggestions to simplify the code. It could also use more testing. Test how this works with multiple codes on the codable concept. It displays in a little bit of a funny way as `code1, code2: desc1, desc2`. Perhaps it should keep the paired codes and descriptions together or just use the first code.

__tests__/utils/fhir.test.ts
```diff
@@ -175,6 +259,31 @@ describe('getDataRequirementFiltersString', () => {
...
+  });
+  test('returns the code and display for resource with both', () => {
+    expect(getFhirResourceSummary(PROCEDURE_RESOURCE_WITH_FULL_SUMMARY)).toEqual(
+      '(123: This is an example of display text for a Procedure resource.)'
+    );
+  });
+  test('returns only the code when the display does not exist but the code does', () => {
+    expect(getFhirResourceSummary(PROCEDURE_RESOURCE_WITH_CODE)).toEqual('(123)');
+  });
+  test('returns only the display when the code does not exist but the code does', () => {
```
sarahmcdougall, line 277:
```suggestion
  test('returns only the display when the code does not exist but the display does', () => {
```

sarahmcdougall (changes requested):

__tests__/utils/fhir.test.ts
```diff
@@ -158,6 +163,85 @@ const TEST_MEASURE_BUNDLE_WITH_EXPANSION: fhir4.Bundle = {
   ]
 };
 
+const RESOURCE_WITH_NO_SUMMARY: fhir4.Resource = {
+  resourceType: 'Prodcedure'
```
hossenlopp, line 167: Misspelling. purposeful or not? It would be good to have a test with a resource without a primary code path entry, ie. MeasureReport, but does have an `id`.

__tests__/utils/fhir.test.ts
```diff
@@ -158,6 +163,85 @@ const TEST_MEASURE_BUNDLE_WITH_EXPANSION: fhir4.Bundle = {
...
+};
+
+const RESOURCE_WITH_NO_CODE_PATH: fhir4.Resource = {
```
hossenlopp, line 170: I think `Procedure` does have a primary code path. So this example has nothing at the primary code path.

__tests__/utils/fhir.test.ts
```diff
@@ -175,6 +259,31 @@ describe('getDataRequirementFiltersString', () => {
   });
 });
 
+describe('getFhirResourceSummary', () => {
+  test('returns an empty string for resource with no code, display, or id', () => {
+    expect(getFhirResourceSummary(RESOURCE_WITH_NO_SUMMARY)).toEqual('');
+  });
+  test('returns the resource id for resource with no primary code path', () => {
+    expect(getFhirResourceSummary(RESOURCE_WITH_NO_CODE_PATH)).toEqual('(procedure-id)');
```
hossenlopp, line 267: See above. This test is a slight misnomer.

hossenlopp (changes requested):
Just a few suggestions with testing.

__tests__/utils/fhir.test.ts
elsaperelli: I added a test with a resource without a primary code path entry.

__tests__/utils/fhir.test.ts
elsaperelli: I kept this and just changed it to test for a resource without a code and I added a different test that tests for a resource without a primary code path.

sarahmcdougall (approved):

hossenlopp (approved):
Looks good!!

PR #27, opened 2022-08-16
Style updates

mgramigna:
# Summary

Makes some low-hanging fruit style updates to the app after our initial UI/UX consultation. Larger in scope changes are going to be separate tasks and may require mockups.

This PR unifies our usage of buttons, adds some small indicators to the user about what is happening, and simplifies some other style-related code

## New Behavior

* Expanded patient now highlights so it is clear what patient is selected
* Left panel is always shown, either with text telling the user what to do or with the data elements
* Data element view now displays text for which patient is being edited
* All icon buttons now have a tooltip that shows up on hover after 1 second explaining what that button is for 
  * The icons themselves are pretty intuitive, but a user can linger if they are trying to get more info. This is a common practice on websites

## Code Changes

* Change all "secondary action" buttons to default variant
* Change population calculation/results buttons to unified size
* Add conditional highlighting of selected patient
* Surround all icon buttons with tooltips
* Remove erroneous CSS
* Change grid widths to not be responsive based on if a patient is expanded
* Add helpful text in data element panel if no patient is selected yet for adding data elements

# Testing Guidance

Test all features of app


elsaperelli (approved):
These are really great styling changes. Looks good to me!

hossenlopp:
image
Peculiar thing... This popup for edit resource stays up after you post it. Might not be something easily resolvable and actually just Mantine's fault.

hossenlopp (approved):

PR #30, opened 2022-08-19
Added confirmation modal to resource delete

elsaperelli:
# Summary
The confirmation modal is now rendered when a user wants to delete a data element.

## New Behavior
When a user clicks the delete button on a data element for a test patient, a confirmation modal appears that prompts the user whether or not they want to delete that data element. The resource type, code, and display text are shown in the modal so that the user knows which element they are deleting, but the text cuts off with ellipses after two lines in case the display text is really long. 

## Code Changes
- `TestResourceCreation.tsx`: code for rendering the confirmation modal when the user clicks the delete button.
- `TestResourceCreation.test.tsx`: unit test for the resource delete confirmation modal.
- `ConfirmationModal.tsx`: some styling changes for the confirmation modal itself so that it works for both patient deletion and resource deletion.

# Testing Guidance
**Functionality:**
1. Upload a measure bundle to fqm-testify and create a test patient.
2. Add some data elements to that test patient.
3. Pick a data element to delete and click the 'trash' icon. A modal should appear that says "Are you sure you want to delete" followed by the resourceType, code, and display text for that data element.
4. Click 'Yes'. The modal should close and that data element should no longer be in the test of test case resources for that patient.
5. Try this a couple times and make sure that every time the correct data element is being displayed in the modal and being deleted. Also make sure that if you click 'cancel' then that data element is not deleted.

**Appearance:**
- Before, the confirmation modal had an AlertTriangle icon in the display message, but it was very difficult to make it a uniform size in a reusable confirmation modal so I decided to take it out since we were likely going to change design elements of the app anyways. With that said, let me know if you think it should still include it or if you have any other design suggestions for the modal!
- I also cut the text in the modal off after two lines because some display string can be very long, but leave a comment if you think it would be more useful for the user to see that entire text, regardless of how long it is.
- Include any other changes you would make in how the confirmation modal is presented!

### DIFF
--- a/components/ResourceCreation/TestResourceCreation.tsx
+++ b/components/ResourceCreation/TestResourceCreation.tsx
@@ -140,11 +140,11 @@ function TestResourceCreation() {
     return undefined;
   };
 
-  const getConfirmationModalText = (id: string | null) => {
-    if (selectedPatient && id) {
-      const resourceIndex = currentTestCases[selectedPatient].resources.findIndex(r => r.id === id);
+  const getConfirmationModalText = (resourceId: string | null) => {
+    if (selectedPatient && resourceId) {
+      const resourceIndex = currentTestCases[selectedPatient].resources.findIndex(r => r.id === resourceId);
       const resource = currentTestCases[selectedPatient].resources[resourceIndex];
-      return `Are you sure you want to delete ${resource.resourceType} ${getFhirResourceSummary(resource)}`;
+      return `Are you sure you want to delete ${resource.resourceType} ${getFhirResourceSummary(resource)}?`;
     }
   };
 
--- a/components/ResourceCreation/TestResourceCreation.tsx
+++ b/components/ResourceCreation/TestResourceCreation.tsx
@@ -10,6 +10,7 @@ import { createFHIRResourceString, getFhirResourceSummary } from '../../util/fhi
 import { selectedPatientState } from '../../state/atoms/selectedPatient';
 import { Edit, Trash } from 'tabler-icons-react';
 import { measurementPeriodState } from '../../state/atoms/measurementPeriod';
+import ConfirmationModal from '../ConfirmationModal';
 
 function TestResourceCreation() {
   const [currentTestCases, setCurrentTestCases] = useRecoilState(patientTestCaseState);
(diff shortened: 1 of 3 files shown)


components/ResourceCreation/TestResourceCreation.tsx
```diff
@@ -115,6 +140,14 @@ function TestResourceCreation() {
     return undefined;
   };
 
+  const getConfirmationModalText = (id: string | null) => {
```
sarahmcdougall, line 143: Nitpick but maybe we should change `id` to `resourceId` so that it's clear that the id refers to the resource and not the selected patient (and to be consistent with the `getConfirmationModalText` in `PatientCreation.tsx` that uses `patientId` as the input rather than `id`)

components/ResourceCreation/TestResourceCreation.tsx
```diff
@@ -115,6 +140,14 @@ function TestResourceCreation() {
...
+    if (selectedPatient && id) {
+      const resourceIndex = currentTestCases[selectedPatient].resources.findIndex(r => r.id === id);
+      const resource = currentTestCases[selectedPatient].resources[resourceIndex];
+      return `Are you sure you want to delete ${resource.resourceType} ${getFhirResourceSummary(resource)}`;
```
sarahmcdougall, line 147: I think it would be good to include a question mark at the end of this string, but not sure if it would look weird with an ellipsis in front of it. Maybe we can play around with this and see how it would look with longer resource displays.

sarahmcdougall (changes requested):
Works great! Aside from the two in-line comments, I wonder if we should consider making the "Yes" button in the confirmation modal red, or changing some other element of the styling, to replace the "alertness" that was previously given by the alert icon. The user may not immediately know that this modal is confirming deletion since the button is blue and nothing else on the modal really stands out. Or we can try outlining the "Yes" button in red or something like that!

components/ResourceCreation/TestResourceCreation.tsx
elsaperelli: image

That's a great idea! I didn't even notice that lol. So this is an example of what it looks like when the display text is really long. There are definitely some even longer ones but they are not as common. Do you think that would be okay/helpful to the user? Or is it too overwhelming?

natjoe4 (approved):
Nice job. Works well

sarahmcdougall (approved):
Thanks for making those changes! Looks good!

PR #32, opened 2022-08-23
Element reference codes

elsaperelli:
# Summary
Updated the `parsePrimaryCodePath.ts` file (now `parseCodePath.ts`) so that it outputs all code paths on each of the resources rather than just the primaryCodePath to `codePaths.ts`,

## New Behavior
- The only new behavior is when `parseCodePaths.ts` is run, it outputs all code paths on each of the resources in `codePaths.ts`. However, `codePaths.ts` is already there so it does not need to be re-run.
- All other behavior should be exactly the same.

## Code Changes
- `package.json`: updated script to run the newly named `parseCodePath.ts` file.
- `parseCodePath.ts`: updated the `parsePrimaryCodePath.ts` file to include all code paths for each resource. 
- `codePaths.ts`: new file outputted by `parseCodePath.ts`
- `fhir.ts`: updated the code that used the output of `parsePrimaryCodePath.ts` to use the output of `parseCodePath.ts`
- `types.ts`: replaced the `PrimaryCodePathInfo` interface with two interfaces, `ResourceCodeInfo` and `CodePathInfo`, that are used in `parseCodePath.ts`

# Testing Guidance
1. `npm run test` to make sure that all the tests still pass and nothing has broken
2. `npm run dev` to make sure that fqm-testify still functions as intended
3. Look at the `codePaths.ts` file and ensure that everything we need is included there.


util/types.ts
```diff
@@ -1,5 +1,8 @@
-export interface PrimaryCodePathInfo {
+export interface ResourceCodeInfo {
   primaryCodePath: string;
-  primaryCodeType?: string;
-  multipleCardinality: boolean;
+  paths: CodePathInfo;
+}
+
+export interface CodePathInfo {
+  [key: string]: { codeType: string; multipleCardinality: boolean; choiceType: boolean };
```
hossenlopp, line 7:
```suggestion
  paths: Record<string, CodePathInfo>;
}

export interface CodePathInfo {
  codeType: string;
  multipleCardinality: boolean;
  choiceType: boolean;
```

hossenlopp (commented):
A few TS niceity suggestions...

hossenlopp (commented):
One last thing. If the primaryCodePath doesn't end up having a code like type, then it doesn't get put in the `paths` record set. So this could be an issue where Questionnaire and Measure would not have info for their primaryCodePath... because it isn't a code like type.

hossenlopp (approved):
After further discussion, only Questionnaire and SearchParameter are the ones with non-'code' like 'primaryCodePaths'. These are incredibly unlikely to be used in measure logic so it's ok that info on those primaryCodePaths is not there.

sarahmcdougall (approved):
The script runs successfully and the information is accurate. Looks like the appropriate changes have been made to reflect the change in script name and script content. Good work!

PR #33, opened 2022-08-30
Element reference codes

elsaperelli:
# Summary
When testing fqm-testify with the "DRCommunicationWithPhysicianManagingDiabetesFHIR-bundle" bundle, Dave noticed that the app crashed and was not able to display data elements with direct reference codes.

## New Behavior
All of the components of the app are the same, but now when a data element doesn't have a valueSet but has a direct reference code, that is what is displayed in the test resource display panel and the scaffolded FHIR resource JSON. In addition, the code and display shown under each test resource for a patient display the direct reference code and display if it exists.

In general, the application now adds all code paths of type FHIR.CodeableConcept, FHIR.Coding, and FHIR.code to the scaffolded FHIR resource JSON. Changes in associated files were made to reflect the changes in parseCodePaths.js.

## Code Changes
- `types.ts` - the structure of the codePaths.ts file was modified so that each dataRequirement object can have more than one path object.
- `parseCodePaths.ts` / `parsePrimaryCodePaths.ts` - modified script that now parses all code type paths for data requirements rather than just the primary code path which is why `parsePrimaryCodePaths.ts` was renamed to `parseCodePaths.ts`. These changes are shown in the previous PR for this with the addition of being able to find elements of type System.string that are embedded in FHIR.Elements. These are just FHIR.code type paths so they are now included in the parse code paths output file.
- `codePaths.ts` / `primaryCodePaths.ts` - `codePaths.ts` replaced `primaryCodePaths.ts` because the new script parses all code paths of all code types into that file. not just the primaryCodePath. The file was renamed to reflect this change.
- `fhir.ts` - functions such as `createFhirResourceString`, `getFhirResourceSummary`, and `getResourcePrimaryCode` were modified to reflect the changes in `codePaths.ts` as well display direct reference codes and add them to the scaffolded Fhir resource JSON.
- `fhir.test.ts` - unit tests updated to reflect changes to `fhir.ts`.
- `package.json` - fixed small mix up.

# Testing Guidance
1. Please ask me questions if any of this doesn't make sense.
2. Run `npm run test` and make sure that all of the tests pass. 
3. Run `npm run dev`
4. Upload the measure bundle "DRCommunicationWithPhysicianManagingDiabetesFHIR-bundle" that is linked on the Jira ticket for this task (#1108). It is also in the slack thread that is linked below. 
5. Create a new patient. Scroll through the available data elements. There are two "Communication" elements that have direct reference codes rather than valuesets. They should look like the following:

image

6. Try uploading other measure bundles, with or without direct reference codes, and make sure nothing breaks and everything looks correct.

### DIFF
--- a/__tests__/utils/fhir.test.ts
+++ b/__tests__/utils/fhir.test.ts
@@ -429,7 +429,6 @@ describe('createFHIRResourceString', () => {
       PERIOD_START,
       PERIOD_END
     );
-    console.log(createdResource);
     expect(Array.isArray(JSON.parse(createdResource).topic)).toBe(true);
     expect(JSON.parse(createdResource).subject).toBeUndefined();
   });
--- a/__tests__/utils/fhir.test.ts
+++ b/__tests__/utils/fhir.test.ts
@@ -25,11 +25,33 @@ const PERIOD_START = '2020-01-01T00:00:00.000Z';
 const PERIOD_END = '2020-12-31T00:00:00.000Z';
 
 const VS_MAP = { testvs: 'test vs name' };
-const DATA_REQUIREMENT_WITH_NO_VALUE_SETS = {
+
+// data requirement that does not have any valuesets or any direct reference codes
+const DATA_REQUIREMENT_WITH_NO_VALUESETS_OR_DRC = {
   type: 'Observation',
   status: 'draft'
 };
-const DATA_REQUIREMENT_WITH_VALUE_SETS = {
+
+// data requirement that does not have any valuesets but has a direct reference code
+const DATA_REQUIREMENT_WITH_NO_VALUESETS_WITH_DRC = {
+  type: 'Communication',
+  status: 'draft',
+  codeFilter: [
+    {
+      path: 'reasonCode',
+      code: [
+        {
+          system: 'http://snomed.info/sct/731000124108',
+          version: 'http://snomed.info/sct/731000124108/version/201709',
+          display: 'test display 2',
+          code: '456'
+        }
+      ]
+    }
+  ]
+};
+
+const DATA_REQUIREMENT_WITH_VALUESETS = {
   type: 'Observation',
   status: 'draft',
   codeFilter: [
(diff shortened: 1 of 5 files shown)


__tests__/utils/fhir.test.ts
```diff
@@ -394,14 +421,15 @@ describe('createFHIRResourceString', () => {
...
-  test('returns populated FHIR resource where primaryCodePath us 0..* or 1..* (multiple cardinality)', () => {
+  test('returns populated FHIR resource where primaryCodePath is 0..* or 1..* (multiple cardinality)', () => {
     const createdResource = createFHIRResourceString(
       ACTIVITYDEFINITION_DATA_REQUIREMENT,
       TEST_MEASURE_BUNDLE,
       'Patient1',
       PERIOD_START,
       PERIOD_END
     );
+    console.log(createdResource);
```
natjoe4, line 432:
```suggestion
```

natjoe4, line 432:
```suggestion
```
👀

natjoe4 (changes requested):
Just one tiny inline comment. Was able to reproduce all described functionality and code looks good! Nice work :)

hossenlopp (approved):
Awesome and robust improvement to what we originally had!

scripts/parseCodePath.ts
```diff
@@ -26,6 +26,8 @@ interface elementChoice {
 export async function parse(xml: string) {
   const { modelInfo } = await xml2js.parseStringPromise(xml);
   const domainInfo = modelInfo.typeInfo.filter((ti: any) => ti.$.baseType === 'FHIR.DomainResource');
+  const elementInfo = modelInfo.typeInfo.filter((ti: any) => ti.$.baseType == 'FHIR.Element' && !ti.$.identifier);
```
natjoe4, line 29:
```suggestion
  const elementInfo = modelInfo.typeInfo.filter((ti: any) => ti.$.baseType === 'FHIR.Element' && !ti.$.identifier);
```
I'm guessing we want === here unless this was intentional?

natjoe4 (changes requested):
Super small change... potentially no change at all

scripts/parseCodePath.ts
elsaperelli: Yes you are right!

natjoe4 (approved):
Nice work!

PR #35, opened 2022-09-02
Upload page

natjoe4:
# Summary
Moved the measureUpload and datePicker components to a separate page

## New Behavior
datePicker and Measure Bundle upload are now on a separate page
A measure must be uploaded (and dates picked) before navigating to test resource creation
The measure Bundle upload page can be returned to by clicking the edit button on the test resource creation page

## Code Changes
Created new `generateTestCases` page
Broke resource creation out of index and added it to generateTestCases Page
Added measure Bundle & measurement period display to generateTestCases page in line with Mockups

# Testing Guidance
- `npm run check`
- `npm run dev`
- Note the `next` button is disabled until a measure is uploaded
- Upload a measure and click next
- Ensure the bundle and period display is accurate and in line with mockup
- Add test resources
- Head back to measure upload by clicking the edit icon
- Head back to test resource creation and ensure the data persists


pages/index.tsx
```diff
@@ -21,9 +26,16 @@ const Home: NextPage = () => {
             <DateSelectors />
           </Grid.Col>
         </Grid>
-        <div style={{ paddingTop: '24px' }}>
-          <TestCaseEditor />
-        </div>
+        <Link href={'generateTestCases'}>
```
mgramigna, line 29: best practice to not have camelCase link names since broswers can't really differentiate. I would suggest either renaming the page to something like `edit`??? or use kebab: `generate-test-cases`. This would involve renaming both the link and the file

mgramigna (changes requested):

pages/index.tsx
mgramigna: I don't _think_ this affects `next build` per se since the error message is quite cryptic, but you never know!

natjoe4: Updated, but no luck on build :(

state/selectors/measurementPeriodString.ts
```diff
@@ -0,0 +1,35 @@
...
+  'April',
+  'May',
+  'June',
+  'July',
+  'August',
+  'September',
+  'October',
+  'November',
+  'December'
+];
```
mgramigna, line 35: ;) 

```suggestion
  return date.toLocaleString('default', {
    month: 'long',
    day: 'numeric',
    year: 'numeric'
  });
};
```

state/selectors/measurementPeriodString.ts
```diff
@@ -0,0 +1,35 @@
+import { selector } from 'recoil';
+import { measurementPeriodState } from '../atoms/measurementPeriod';
+
+export const measurementPeriodStringState = selector<string>({
```
mgramigna, line 4: This is a nitpick, but I feel like, conceptually, this doesn't really need to be a selector. It's just a string that is rendered based on the values of another atom, however the rest of the app doesn't need access to it. It just renders once at the top. It might make sense to move the date string logic into the actual component that renders the banner. Your call though

pages/generate-test-cases.tsx
```diff
@@ -0,0 +1,52 @@
...
+              </div>
+            </Card>
+            <Text color="dimmed" size="xl">
+              FQM Testify: an eCQM Analysis Tool
+            </Text>
+          </div>
+        </Grid.Col>
+      </Grid>
+      <div style={{ paddingTop: '24px' }}>
+        <TestCaseEditor />
```
mgramigna, line 48: Theoretically, someone could navigate to this page by entering the URL directly rather than through clicking the buttons. It might be a good idea to render a helpful message and a prompt for navigating back to the homepage, instead of the test case editor.

Even just some centered text with like a button or something

pages/generate-test-cases.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <Grid>
+        <Grid.Col span={12} sx={() => ({ backgroundColor: 'black', height: 'auto', padding: 20 })}>
+          <div
+            style={{
+              display: 'flex',
+              justifyContent: 'space-between',
+              alignItems: 'center'
+            }}
+          >
+            <Card sx={() => ({ width: 'fit-content' })}>
```
mgramigna, line 24: Similar to comment about how this page _could_ be navigated to without a measure bundle (only if the user goes awry), maybe we only render this card if the measure bundle/period is not null. This goes in tandem with adding helpful text instead of the test case editor for navigating back to the homepage

mgramigna (changes requested):

pages/generate-test-cases.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { measureBundleState } from '../state/atoms/measureBundle';
+import { measurementPeriodStringState } from '../state/selectors/measurementPeriodString';
+import { Edit } from 'tabler-icons-react';
+
+export default function TestCaseEditorPage() {
+  const measurementPeriodString = useRecoilValue(measurementPeriodStringState);
+  const measureBundle = useRecoilValue(measureBundleState);
+  return (
+    <>
+      <Grid>
```
mgramigna, line 15: We should probably just yoink this and replace the `AppHeader` component? And if the card is only rendered when the measurement period is populated, that's nice? what are your thoughts?

mgramigna (approved):
looks good

elsaperelli (approved):
Nice! Looks great

PR #36, opened 2022-09-07
Highlighting update

elsaperelli:
# Summary
The measure highlighting for individual patients now automatically updates when the patient data changes rather than the user having to click the calculate button each time.

## New Behavior
- There is no more calculate button in `MeasureHighlightingPanel.tsx` because the measure highlighting automatically recalculates when there is a change to the patient data. Instead, there is an loading icon that shows when the measure highlighting is being recalculated, or a check mark that shows that the highlighting is up to date. 
- When a user selects a patient, the measure highlighting for that patient immediately appears, even if the patient does not have any resources. When a user adds resources, the highlighting updates to reflect those changes. The highlighting also updates when the patient data is edited, when resources are added or edited, and when resources are deleted.
- Since the measure highlighting also changes when the measure bundle or measure period changes, the measureReport calculation is reran every time the user navigates to the TestCaseEditorPage. This is how the measure highlighting initially shows up and how the measure report is recalculated if the user navigates back to the measure upload page and uploads a different bundle or selects a different measurement period.

## Code Changes
- `MeasureCalculation.tsx` - this helper function/file was created to hold a function that takes in a patientTestCase, a measure bundle, and a measurement period start and end and returns a promise for a MeasureReport. This is in place of the function that formerly resided in `MeasureHighlightingPanel.tsx`.
- `MeasureHighlightingPanel.tsx` - the calculate function was removed and replaced with a call to the `calculateMeasureReport` helper function and the calculate button was also removed. Icons were added to the panel to indicate when the panel was loading or up-to-date.
- `PatientCreationPanel.tsx` - calls to the `calculateMeasureReport` helper function are added where the patient data changes. 
- `ResourceDisplay.tsx` - calls to the calculateMeasureReport` helper function are added where the patient data changes. 
- `generate-test-cases.tsx`- a function is added and called in order to recalculate the each patients' measure reports when the user navigates to the `TestCaseEditorPage`. Also, 
- `calculationLoading.ts` - an atom indicating whether or not measure reports are being recalculated was added. 

# Testing Guidance
1. Run `npm run test` to make sure that all of the tests pass.
2. Run `npm run dev` 
3. Upload a measure bundle and click next. Create a patient. The measure highlighting for that patient should immediately render.
4. Add a resource to the patient. The measure highlighting should update to reflect that change.
5. Try out a bunch of different use cases. Add resources, delete resources, edit the patient, upload a new measure bundle. Make sure that the measure highlighting updates properly each time.

### DIFF
--- a/components/resource-creation/ResourceDisplay.tsx
+++ b/components/resource-creation/ResourceDisplay.tsx
@@ -119,12 +119,10 @@ function ResourceDisplay() {
             draftState[selectedPatient].resources[resourceIndexToUpdate] = updatedResource;
           }
           // re-run measure report calculations for updated state
-          measureReportCalculation(draftState, selectedPatient);
+          await measureReportCalculation(draftState, selectedPatient);
         }).then(nextResourceState => {
-          setTimeout(() => {
-            setCurrentTestCases(nextResourceState);
-            setIsCalculationLoading(false);
-          }, 2000);
+          setCurrentTestCases(nextResourceState);
+          setIsCalculationLoading(false);
         });
       }
     }
@@ -139,7 +137,7 @@ function ResourceDisplay() {
         produce(currentTestCases, async draftState => {
           draftState[selectedPatient].resources.splice(resourceIndexToDelete, 1);
           // re-run measure report calculations for updated state
-          measureReportCalculation(draftState, selectedPatient);
+          await measureReportCalculation(draftState, selectedPatient);
         }).then(nextResourceState => {
           setCurrentTestCases(nextResourceState);
           setIsCalculationLoading(false);
--- a/components/resource-creation/ResourceDisplay.tsx
+++ b/components/resource-creation/ResourceDisplay.tsx
@@ -5,16 +5,17 @@ import { useRecoilState, useRecoilValue, useSetRecoilState } from 'recoil';
 import CodeEditorModal from '../modals/CodeEditorModal';
 import { measureBundleState } from '../../state/atoms/measureBundle';
 import { selectedDataRequirementState } from '../../state/atoms/selectedDataRequirement';
-import { patientTestCaseState } from '../../state/atoms/patientTestCase';
+import { patientTestCaseState, TestCase } from '../../state/atoms/patientTestCase';
 import { createFHIRResourceString, getFhirResourceSummary } from '../../util/fhir';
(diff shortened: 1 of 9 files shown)


components/resource-creation/ResourceDisplay.tsx
```diff
@@ -103,10 +130,33 @@ function ResourceDisplay() {
     if (id && selectedPatient) {
       const resourceIndexToDelete = currentTestCases[selectedPatient].resources.findIndex(r => r.id === id);
       if (resourceIndexToDelete >= 0) {
-        const nextResourceState = produce(currentTestCases, draftState => {
+        setIsCalculationLoading(true);
+        produce(currentTestCases, async draftState => {
           draftState[selectedPatient].resources.splice(resourceIndexToDelete, 1);
+          // re-run measure report calculations for updated state
+          if (measureBundle.content) {
```
natjoe4, line 137: it seems like this conditional w/ try/catch is used multiple times. Can we break that out into a function?

natjoe4 (changes requested):
Nice job! I think in an ideal world, the loading icon would be centered in the highlight panel instead of stuck at the top where it can leave view. What do you think? If this change is easy, it could be worth doing.

util/MeasureCalculation.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Takes in a patient's test case info, a measure bundle, and a measurement
+ * period start and end and returns a measureReport
+ * passed in filename and fileContents
+ * @param patientTestCase {TestCaseInfo}
+ * @param mb {fhir4.Bundle}
+ * @param mpStart {string}
+ * @param mpEnd {string}
+ * @returns MeasureReport
```
mgramigna, line 13: This is opinionated, but one of the reasons TypeScript is so good is that it basically eliminates the need for JSDoc comments like this. All of this info is clearly identifiable by just looking at the function signature, so my opinion is that we shouldn't even put waste time writing these.

natjoe4 (approved):
Nice job! Works great and looks good too.

natjoe4 (approved):
Nice! I like that the spinner always appears now

hossenlopp (approved):
Woo! Works great and is clear when the calculation is updated.

natjoe4 (approved):
🎉

hossenlopp (approved):

PR #38, opened 2022-09-12
Add ability to create resource not from data requirements

sarahmcdougall:
# Summary
Adds the ability for the user to add any FHIR resource to a given test case so that the user is not restricted to only the resources derived from the data requirements.

## New Behavior
The user now has the option to select a resource (derived from the data requirements) from the select component, or to add a new resource to the test case by pasting the resource JSON in a blank modal that will create the new resource in the resource panel. 

## Code Changes
* `resource-creation/ResourcePanel`
    * New button “Add New Custom Resource” that will bring up the blank modal for inputting FHIR JSON data
    * New function `createNewResource` to create the new FHIR resource from the JSON and add it to the selected patient (Note: right now, the function ensures that the new resource has an id since that is needed for editing the resource from the ResourceInfoCard. In the future, we will need to validate the incoming JSON as FHIR. Is there anything else that should be checked as of now like resource type or any other important fields?)
* `resource-creation/ResourceSelection`
    * Changed the placeholder text on the select dropdown so that the two resource creation methods are more clear

# Testing Guidance
* After uploading a measure bundle and selecting a patient, try to add a new resource by clicking on the “Add New Custom Resource” button
* Input JSON into the modal and click Save
    * Try a FHIR resource from an EXM bundle, an empty JSON object, a JSON object without a resource type, etc. to try and see if anything breaks or if any errors should be thrown for missing information
    * Try all the resource info functionality with this new resource once it is created - deleting it, editing it, etc.
* Run the unit tests

### DIFF
--- a/components/resource-creation/ResourcePanel.tsx
+++ b/components/resource-creation/ResourcePanel.tsx
@@ -1,6 +1,6 @@
 import { Button, Center, Divider, Text } from '@mantine/core';
 import { v4 as uuidv4 } from 'uuid';
-import { IconCodePlus } from '@tabler/icons';
+import { IconAlertCircle, IconCodePlus } from '@tabler/icons';
 import { Suspense, useState } from 'react';
 import { useRecoilState, useRecoilValue } from 'recoil';
 import { Loader } from 'tabler-icons-react';
@@ -11,6 +11,7 @@ import produce from 'immer';
 import CodeEditorModal from '../modals/CodeEditorModal';
 import ResourceDisplay from './ResourceDisplay';
 import ResourceSelection from './ResourceSelection';
+import { showNotification } from '@mantine/notifications';
 
 export default function ResourcePanel() {
   const selectedPatient = useRecoilValue(selectedPatientState);
@@ -25,10 +26,21 @@ export default function ResourcePanel() {
       if (!newResource.id) {
         newResource.id = uuidv4();
       }
-      const nextResourceState = produce(currentPatients, draftState => {
-        draftState[selectedPatient].resources.push(newResource);
-      });
-      setCurrentPatients(nextResourceState);
+      const resourceIndex = currentPatients[selectedPatient].resources.findIndex(r => r.id === newResource.id);
+      if (resourceIndex >= 0) {
+        showNotification({
+          id: 'failed-upload',
+          icon: <IconAlertCircle />,
+          title: 'Resource Creation Failed',
+          message: `Resource ID ${newResource.id} is used for an existing resource for this patient.`,
+          color: 'red'
+        });
+      } else {
+        const nextResourceState = produce(currentPatients, draftState => {
+          draftState[selectedPatient].resources.push(newResource);
+        });
+        setCurrentPatients(nextResourceState);
+      }
     }
     setIsNewResourceModalOpen(false);
   };
@@ -70,7 +82,7 @@ export default function ResourcePanel() {
           }}
         >
           <ResourceSelection />
-          <Center>Or</Center>
+          <Center>or</Center>
           <Center>
             <Button
               aria-label="Add New Custom Resource"
--- a/components/resource-creation/ResourcePanel.tsx
+++ b/components/resource-creation/ResourcePanel.tsx
@@ -1,16 +1,37 @@
-import { Center, Divider, Text } from '@mantine/core';
-import { Suspense } from 'react';
-import { useRecoilValue } from 'recoil';
+import { Button, Center, Divider, Text } from '@mantine/core';
+import { v4 as uuidv4 } from 'uuid';
+import { IconCodePlus } from '@tabler/icons';
+import { Suspense, useState } from 'react';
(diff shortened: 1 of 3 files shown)


components/resource-creation/ResourcePanel.tsx
```diff
@@ -1,16 +1,49 @@
...
+      }
+      const resourceIndex = currentPatients[selectedPatient].resources.findIndex(r => r.id === newResource.id);
+      if (resourceIndex >= 0) {
+        showNotification({
+          id: 'failed-upload',
+          icon: <IconAlertCircle />,
+          title: 'Resource Creation Failed',
+          message: `Resource ID ${newResource.id} is used for an existing resource for this patient.`,
+          color: 'red'
+        });
```
elsaperelli, line 37: This was a really good call! I don't know if this is in your scope of the task, but I noticed through this that if you add a resource (not a custom one) and then change its ID then it replaces the resource that had that ID with the new one...maybe we should add this notification to that case too!

components/resource-creation/ResourcePanel.tsx
```diff
@@ -1,16 +1,49 @@
...
+      } else {
+        const nextResourceState = produce(currentPatients, draftState => {
+          draftState[selectedPatient].resources.push(newResource);
+        });
+        setCurrentPatients(nextResourceState);
+      }
+    }
+    setIsNewResourceModalOpen(false);
```
mgramigna, line 45: This logic always closes the modal even in the error case, making the user lose their progress if all they needed to do was change the ID. Maybe we should move this line into the else case above so the modal closes only if the state is actually about to be updated? Otherwise they can correct their error and save again

mgramigna (changes requested):
Looks great, one small change

components/resource-creation/ResourcePanel.tsx
```diff
@@ -43,6 +82,17 @@ export default function ResourcePanel() {
           }}
         >
           <ResourceSelection />
+          <Center>or</Center>
+          <Center>
+            <Button
+              aria-label="Add New Custom Resource"
+              onClick={() => setIsNewResourceModalOpen(true)}
+              variant="outline"
+            >
```
elsaperelli, line 91: Would you oppose making the "Add New Custom Resource" option as an option in the dropdown menu rather than a button? Might make it look more similar to the original design but I also like the button!

elsaperelli (changes requested):
Great job! Just one comment. Also, in terms of if there is anything else we need to check in the code editor modal, I don't think so since its; the same as if the code was being edited.

components/resource-creation/ResourcePanel.tsx
sarahmcdougall: That's a good point. Will there ever be a case where we have a resource with, for example, id `123` and then we edit a different resource to have id `123` but we actually want the second resource to replace the first one (in terms of references, etc.)? If not, then I think we can easily add the notification for this case as well. @mgramigna thoughts?

components/resource-creation/ResourcePanel.tsx
sarahmcdougall: I prefer the button to a dropdown menu option so that we can more easily differentiate the two options for resource creation (from requirements versus from manually adding a resource with a modal). I think it would be more difficult to differentiate the two options if we had it in the dropdown because then we would need to also somehow make it clear that the other resources came from the data requirements

components/resource-creation/ResourcePanel.tsx
mgramigna: in general I don't think users will ever really be editing IDs. I'm fine with restricting behavior such that IDs need to be unique during any creation of a resource in a modal (no overwriting. do that through the edit functionality if you want to replace a resource)

you _can_ make that change here to show an error in the case Elsa mentioned too but you don't have to

sarahmcdougall: After looking at the code more, I think I will hold off on making the change here since creating a new resource and updating an existing resource both call `updateResource` (but follow different branches within the function depending on whether the id already exists for the current patient). I think making this change would require some larger changes to determine whether we are creating a new resource with an existing id because that would mean we would need to pass in information aside from the id saying that this is a new resource (not sure if that makes sense sorry - I'm looking at line 75 in ResourceDisplay.tsx where we determine whether to update/create based on whether the id is already present)

sarahmcdougall: After looking at the code more, I think I will hold off on making the change here since creating a new resource and updating an existing resource both call updateResource (but follow different branches within the function depending on whether the id already exists for the current patient). I think making this change would require some larger changes to determine whether we are creating a new resource with an existing id because that would mean we would need to pass in information aside from the id saying that this is a new resource (not sure if that makes sense sorry - I'm looking at line 75 in ResourceDisplay.tsx where we determine whether to update/create based on whether the id is already present)

mgramigna (approved):

elsaperelli (approved):
Looks awesome!

PR #40, opened 2022-09-15
MB upload information added to readme and app

elsaperelli:
# Summary
The README as well as the app now include more information on what the uploaded measure bundle should contain as well as instructions on how to add ValueSets to a measure bundle through the fqm-execution CLI. 

## New Behavior
- The first page of the app where the user uploads a measure bundle now includes a sub-header to the main `AppHeader` called the `MeasureUploadHeader`. This will contain more information in the future, but for now it says "Step 1: Upload a Measure Bundle" as well as a popover button next to that text that displays more information about the measure bundle when clicked. 

## Code Changes
- `README.md` - specifications for what the uploaded measure bundle must contain and instructions on how to use the fqm-execution in order to add ValueSets to a measure bundle.
- `MeasureUploadHeader.tsx` - a sub-header component that is only displayed on the first page of the app and contains the measure upload information popover.
- `index.tsx` - renders the `MeasureUploadHeader` on the homepage.
- `MeasureUploadHeader.test.tsx` - tests that the header title and information popover are rendered.

# Testing Guidance
1. Run `npm run test`
2. Make sure that the `MeasureUploadHeader` is only rendered on the first page
3. Check spelling, grammar, missing information etc. in the `README.md` as well as the information popover in the app.
4. Let me know if you have any ideas or suggestions on anything!

### DIFF
--- a/components/utils/MeasureUploadHeader.tsx
+++ b/components/utils/MeasureUploadHeader.tsx
@@ -26,8 +26,10 @@ export default function MeasureUploadHeader() {
             <List.Item>All dependent library resources used</List.Item>
             <List.Item>
               All ValueSets needed for measure calculation (see{' '}
-              <Anchor href="https://github.com/projecttacoma/fqm-testify#adding-test-cases">here</Anchor> for more info
-              on how to obtain necessary ValueSets)
+              <Anchor href="https://github.com/projecttacoma/fqm-testify#adding-valuesets-to-fhir-measure-bundle">
+                here
+              </Anchor>{' '}
+              for more info on how to obtain necessary ValueSets)
             </List.Item>
           </List>
         </Popover>
--- a/components/utils/MeasureUploadHeader.tsx
+++ b/components/utils/MeasureUploadHeader.tsx
@@ -22,7 +22,7 @@ export default function MeasureUploadHeader() {
         >
           The uploaded Measure Bundle must contain:
           <List>
-            <List.Item>One measure resource</List.Item>
+            <List.Item>Exactly one measure resource</List.Item>
             <List.Item>All dependent library resources used</List.Item>
             <List.Item>
               All ValueSets needed for measure calculation (see{' '}
--- a/components/utils/MeasureUploadHeader.tsx
+++ b/components/utils/MeasureUploadHeader.tsx
@@ -1,4 +1,4 @@
-import { Text, Popover, List, Anchor, ActionIcon } from '@mantine/core';
+import { Text, Popover, List, Anchor, ActionIcon, Group } from '@mantine/core';
 import React, { useState } from 'react';
 import { InfoCircle } from 'tabler-icons-react';
 
@@ -6,15 +6,7 @@ export default function MeasureUploadHeader() {
   const [opened, setOpened] = useState(false);
 
   return (
-    <div
-      style={{
-        display: 'flex',
-        alignItems: 'center',
-        height: '100%',
-        width: '100%',
-        padding: 20
-      }}
-    >
+    <Group>
       <div style={{ paddingRight: 5 }}>
         <Text size="xl">Step 1: Upload a Measure Bundle</Text>
       </div>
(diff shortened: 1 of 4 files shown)


components/utils/MeasureUploadHeader.tsx
```diff
@@ -0,0 +1,45 @@
...
+            </ActionIcon>
+          }
+        >
+          The uploaded Measure Bundle must contain:
+          <List>
+            <List.Item>one measure resource</List.Item>
+            <List.Item>all dependent library resources used</List.Item>
+            <List.Item>
+              all ValueSets needed for measure calculation see{' '}
+              <Anchor href="https://github.com/projecttacoma/fqm-testify#adding-test-cases">here</Anchor> for more info
```
mgramigna, line 37: This links to the fqm-testify README for adding test cases. I assumed you wanted a link to fqm-execution instead?

components/utils/MeasureUploadHeader.tsx
```diff
@@ -0,0 +1,45 @@
...
+              <InfoCircle size={20} />
+            </ActionIcon>
+          }
+        >
+          The uploaded Measure Bundle must contain:
+          <List>
+            <List.Item>one measure resource</List.Item>
+            <List.Item>all dependent library resources used</List.Item>
+            <List.Item>
+              all ValueSets needed for measure calculation see{' '}
```
mgramigna, line 36: Can we have the "see ...." be in parentheses?

components/utils/MeasureUploadHeader.tsx
```diff
@@ -0,0 +1,45 @@
...
+          onClose={() => setOpened(false)}
+          target={
+            <ActionIcon aria-label={'More Information'} onClick={() => setOpened(o => !o)}>
+              <InfoCircle size={20} />
+            </ActionIcon>
+          }
+        >
+          The uploaded Measure Bundle must contain:
+          <List>
+            <List.Item>one measure resource</List.Item>
```
mgramigna, line 33: Nitpick but can we capitalize the first word of each list item?

components/utils/MeasureUploadHeader.tsx
```diff
@@ -0,0 +1,45 @@
+import { Text, Popover, List, Anchor, ActionIcon } from '@mantine/core';
+import React, { useState } from 'react';
+import { InfoCircle } from 'tabler-icons-react';
+
+export default function MeasureUploadHeader() {
+  const [opened, setOpened] = useState(false);
+
+  return (
+    <div
```
mgramigna, line 9: Is this div with these styles really necessary? This seems like exactly what a Mantine `Group` component would do. Maybe we can replace it with that?

mgramigna (changes requested):

components/utils/MeasureUploadHeader.tsx
elsaperelli: Ah so I guess it won't appear when you are testing because I made the changes to the README in this PR but it will link to that section and it will then include a section on adding ValueSets to the measure bundle. Does that make sense?

components/utils/MeasureUploadHeader.tsx
elsaperelli: Nice catch

components/utils/MeasureUploadHeader.tsx
mgramigna: Got it. You probably want it to link to `#adding-valuesets-to-fhir-measure-bundle` then so it goes directly to that section. It's fine if you push up a link that doesn't 100% work yet since it'll work when this is merged

mgramigna (approved):

components/utils/MeasureUploadHeader.tsx
```diff
@@ -0,0 +1,39 @@
...
+          onClose={() => setOpened(false)}
+          target={
+            <ActionIcon aria-label={'More Information'} onClick={() => setOpened(o => !o)}>
+              <InfoCircle size={20} />
+            </ActionIcon>
+          }
+        >
+          The uploaded Measure Bundle must contain:
+          <List>
+            <List.Item>One measure resource</List.Item>
```
natjoe4, line 25:
```suggestion
            <List.Item>Exactly one measure resource</List.Item>
```

natjoe4 (changes requested):
Nice job. Really splitting hairs at this point, but I think this is helpful, since more than one measure will also throw an error.

natjoe4 (approved):
This rules

PR #41, opened 2022-09-15
Fix select box not blurring on modal close

mgramigna:
# Summary

Fixes a bug where the resource selection dropdown remained open after closing the resource modal (via either "Save" or "Cancel").

This ended up being an issue with the `searchable` prop. For whatever reason, the combo of using `value=""` and `searchable` caused this bug. The workaround I went with for this is to only enable the searchable prop when the dropdown is open.

## New Behavior

Dropdown stays closed when modal closes

## Code Changes

* Use local state variable for whether or not to enable searching in the dropdown (flips to true when the dropdown is opened by the user)
* Update unit test to use a click instead of a focus to ensure the above functionality triggers

# Testing Guidance

* Add resource via dropdown (make sure you can still search)
* When either canceling or saving the resource, the dropdown should stay closed

### DIFF
--- a/__tests__/components/resource-creation/ResourceSelection.test.tsx
+++ b/__tests__/components/resource-creation/ResourceSelection.test.tsx
@@ -154,7 +154,7 @@ describe('ResourceSelection', () => {
     const resourceSelector = screen.getByPlaceholderText(/select fhir resource/i) as HTMLInputElement;
 
     await act(async () => {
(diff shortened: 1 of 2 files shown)


elsaperelli (approved):
Works great! Truly life changing, actually.

sarahmcdougall (approved):
Arguably the most important task in this sprint. Fixes the issue and doesn't break anything 👍

PR #43, opened 2022-09-16
Measure report state

elsaperelli:
# Summary
Before, the measure report for a patient was a property on the patient's state. Now, we can access the measure report for the patient by accessing it by a different state. 

## New Behavior
There is now a new atom for measureReportLookup, so all of the places where calculation takes place were updated to use the new state, but all other behavior remains the same.

## Code Changes
- `measureReportLookup.ts` - creates a new atom for measureReportLookup.
- `MeasureHighlightingPanel.tsx` - Now parses the measure report that is in the measureReportLookup atom.
- `PatientCreationPanel.tsx` - Updated all the places where calculation takes place to use the same state.
- `ResourceDisplay.tsx` - Updated all the places where calculation takes place to use the same state. Also gets rid of the `setTimeout()`.
- `ResourcePanel.tsx` - Updated all the places where calculation takes place to use the same state. 
- `generate-test-cases.ts` - Updated all the places where calculation takes place to use the same state.

# Testing Guidance
1. Run `npm run check` to ensure that all tests still pass and that there are no lint or prettier errors. 
2. Make sure that the measure highlighting changes in the following cases:
- The measurement period is edited
- The measure bundle is changed
- A test case is imported
- A test case is edited
- A test resource is added to a test case
- A test resource is edited
- A test resource is deleted

### DIFF
--- a/components/patient-creation/PatientCreationPanel.tsx
+++ b/components/patient-creation/PatientCreationPanel.tsx
@@ -119,30 +119,32 @@ function PatientCreationPanel() {
       setCurrentPatients(nextPatientState);
       setIsCalculationLoading(true);
 
-      produce(measureReportLookup, async draftState => {
-        if (measureBundle.content) {
-          try {
-            draftState[patientId] = await calculateMeasureReport(
-              nextPatientState[patientId],
-              measureBundle.content,
-              measurementPeriod.start?.toISOString(),
-              measurementPeriod.end?.toISOString()
-            );
-          } catch (error) {
-            if (error instanceof Error) {
-              showNotification({
-                icon: <IconAlertCircle />,
-                title: 'Calculation Error',
-                message: error.message,
-                color: 'red'
-              });
+      setTimeout(() => {
+        produce(measureReportLookup, async draftState => {
+          if (measureBundle.content) {
+            try {
+              draftState[patientId] = await calculateMeasureReport(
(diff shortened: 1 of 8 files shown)


natjoe4 (changes requested):
Works great! A couple small fixes. 
 - On patient deletion, we should also delete the MeasureReport associated with that patient
 - MeasureReport field can be removed from patientTestCase state

natjoe4 (approved):
🎉

components/patient-creation/PatientCreationPanel.tsx
```diff
@@ -106,17 +109,21 @@ function PatientCreationPanel() {
...
+      const nextPatientState = produce(currentPatients, draftState => {
         draftState[patientId] = {
           patient: pt,
           resources: resources
         };
+      });
+      setCurrentPatients(nextPatientState);
+      setIsCalculationLoading(true);
+
+      produce(measureReportLookup, async draftState => {
```
hossenlopp, line 122: Make this happen in a setTimeout.

hossenlopp (changes requested):

hossenlopp (approved):
Looks good!

natjoe4 (approved):
Sweet

hossenlopp (approved):
Good to go now! :)

PR #46, opened 2022-09-20
Measure upload UI

elsaperelli:
# Summary
This PR updates the UI of the measure upload page to be more inline with that we were given by the UI/UX team in the mockups. There are now "Step 1" and "Step 2" headers for the measure upload and the date pickers. The measure upload and date pickers also take up less room as shown in the mockups. Calendar icons are added to the date pickers.

## New Behavior
There is no new behavior.

## Code Changes
- `pages/index.tsx` - now renders the `DateSelectorsHeader` in addition to the other components on the page. The components are now in a stack so that they can be in a `Grid.Col` has has a span of 5.
- `components/utils/DateSelectorsHeader.tsx` - creates a new `DateSelectorsHeader` component.
- `components/measure-upload/MeasureUpload.tsx` - Gives the drop zone a minimum height so that it is a bit larger on the page.
- `components/measure-upload/DateSelectors.tsx` - Adds a calendar icon to the date pickers and makes the date pickers smaller.

# Testing Guidance
- Run `npm run check` to ensure that all tests pass and there are no warnings.
- Run `npm run dev` to ensure that the app functions the same way (upload a measure, change the dates, etc.)
- Look at the mockup page (attached below) to see that the current design resembles that.
- Give me any feedback about the design!

NOTE: In the mockups, the calendar icon is on the right of the date pickers. It seems as though Mantine only gives  you the option to put it on the left. If you feel strongly about it being on the right, let me know and I will do my best to work around that!

Mockup: 
image

### DIFF
--- a/components/measure-upload/DateSelectors.tsx
+++ b/components/measure-upload/DateSelectors.tsx
@@ -1,5 +1,6 @@
 import { Grid } from '@mantine/core';
 import { DatePicker } from '@mantine/dates';
+import { IconCalendar } from '@tabler/icons';
 import { useRecoilState } from 'recoil';
 import { measurementPeriodState } from '../../state/atoms/measurementPeriod';
 
@@ -7,20 +8,22 @@ export default function DateSelectors() {
   const [period, setPeriod] = useRecoilState(measurementPeriodState);
   return (
     <Grid>
-      <Grid.Col span={6}>
+      <Grid.Col span={5}>
         <DatePicker
           placeholder="Select start date"
           label="Measurement Period Start"
           value={period.start}
           onChange={start => setPeriod({ ...period, start })}
+          icon={<IconCalendar size={16} />}
         />
       </Grid.Col>
-      <Grid.Col span={6}>
+      <Grid.Col span={5}>
         <DatePicker
           placeholder="Select end date"
           label="Measurement Period End"
           value={period.end}
           onChange={end => setPeriod({ ...period, end })}
+          icon={<IconCalendar size={16} />}
         />
       </Grid.Col>
     </Grid>
--- a/components/measure-upload/DateSelectors.tsx
+++ b/components/measure-upload/DateSelectors.tsx
@@ -14,7 +14,7 @@ export default function DateSelectors() {
           label="Measurement Period Start"
           value={period.start}
           onChange={start => setPeriod({ ...period, start })}
-          icon={<IconCalendar size={16} />}
+          icon={<IconCalendar size={25} />}
         />
       </Grid.Col>
       <Grid.Col span={5}>
@@ -23,7 +23,7 @@ export default function DateSelectors() {
           label="Measurement Period End"
           value={period.end}
           onChange={end => setPeriod({ ...period, end })}
-          icon={<IconCalendar size={16} />}
+          icon={<IconCalendar size={25} />}
         />
       </Grid.Col>
     </Grid>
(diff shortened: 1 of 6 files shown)


pages/index.tsx
```diff
@@ -17,27 +18,29 @@ const Home: NextPage = () => {
...
+            <DateSelectors />
+            <Link href={'/generate-test-cases'}>
+              <Group position="right">
+                <Button
+                  sx={() => ({
+                    marginTop: 10
+                  })}
+                  disabled={!(measureBundle.name && measurementPeriod.start && measurementPeriod.end)}
+                >
+                  Next
```
sarahmcdougall, line 37: Should we still be allowed to actually click on the "Next" button and navigate to a new page? If I do not upload a measure bundle, the "Next" button gets disabled but I can still click on it (and then get redirected to the page about not having a measure bundle selected). I know that you can't do anything with the other disabled buttons (in the patient creation panel) so I feel like the user shouldn't be able to go to a new page when this button is disabled either? Just wanted to check if you experience the same behavior

sarahmcdougall (changes requested):
I really like the look of this! It looks a lot cleaner now that we have fixed the spacing/sizing of things.

This is a superrr nitpick but it would be nice if we could have the same spacing between "Step 1" and the drop zone, and "Step 2" and the date pickers (pointed out in the image below). But the spacing isn't that different so I feel like most people won't notice it, so maybe only try and fix this if it's an easy fix and don't worry about it otherwise. They might actually be the same and my eyes are deceiving me because of the outline of the drop zone, not really sure
Screen Shot 2022-09-21 at 10 43 02 AM

package-lock.json
```diff
@@ -6811,7 +6811,7 @@
         "object-keys": {
           "version": "0.4.0",
           "resolved": "https://registry.npmjs.org/object-keys/-/object-keys-0.4.0.tgz",
-          "integrity": "sha512-ncrLw+X55z7bkl5PnUvHwFK9FcGuFYo9gtjws2XtSzL+aZ8tm830P60WJ0dSmFVaSalWieW5MD7kEdnXda9yJw=="
+          "integrity": "sha1-KKaq50KN0sOpLz2V8hM13SBOAzY="
```
mgramigna, line 6814: https://github.com/projecttacoma/fqm-testify/pull/45#discussion_r976515244

components/measure-upload/DateSelectors.tsx
```diff
@@ -1,26 +1,29 @@
...
         />
       </Grid.Col>
-      <Grid.Col span={6}>
+      <Grid.Col span={5}>
         <DatePicker
           placeholder="Select end date"
           label="Measurement Period End"
           value={period.end}
           onChange={end => setPeriod({ ...period, end })}
+          icon={<IconCalendar size={16} />}
```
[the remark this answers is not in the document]
elsaperelli: I agree! Good call :)

pages/index.tsx
elsaperelli: Nice catch!

lmd59 (commented):
Looks pretty good to me! Two small suggestions:
1. Maybe we can make "Step 1" and "Step 2" bold or a different color or something? Looks a little run-on to me, and I thought that might help highlight the step 1/step 2 process.
2. I know folks aren't likely to be using this on a very small screen, but it's weird to me that it starts smooshing things out of their boundaries when there's still plenty of horizontal space.
Screen Shot 2022-09-21 at 3 26 25 PM

sarahmcdougall (approved):
Looks great and works great! :)

PR #47, opened 2022-09-26
UI bug

elsaperelli:
# Summary
This was not a task but just an end of sprint activity. The download, edit, copy, and delete buttons of the `PatientInfoCard` were a little off and it got cut off if the user resized the window. This is how it looked: 

Screen Shot 2022-09-26 at 12 22 41 PM

Screen Shot 2022-09-26 at 12 22 57 PM

 I fixed the layout so that doesn't happen any more.

## New Behavior
- When a user shrinks the screen to a point where the patient name and buttons would overlap, now a menu button appears in place of the four buttons. When the user clicks on the menu button, a dropdown menu appears with those 4 options that are their own buttons when the screen is large enough. All other behavior remains the same.

## Code Changes
- `PatientCreationPanel.tsx` - set `isSmallScreen` to true if the max-width of the window is 1600px.
- `PatientInfoCard.tsx` - if the screen is small, a menu button is rendered in place of the four buttons. 
- `PatientInfoCard.test.tsx` - updated tests to reflect new changes.

# Testing Guidance
1. Run `npm run check` and make sure all tests pass and there are no lint or prettier errors.
2. Run `npm run dev`, upload a measure bundle, and create a patient.
3. In a normal sized window, you should see the four original buttons and there should be no overlap like there was previously.
4. Try resizing the window. If you shrink the window, you should see that the four buttons become one menu button. Try out all of the options in the menu to make sure that they work.
5. Let me know if you find anything or have any style changes/ideas!

### DIFF
--- a/__tests__/components/utils/PatientInfoCard.test.tsx
+++ b/__tests__/components/utils/PatientInfoCard.test.tsx
@@ -169,17 +169,7 @@ describe('PatientInfoCard', () => {
 
   it('should render a patient actions menu when the screen is small', () => {
     render(
-      mantineRecoilWrap(
-        <PatientInfoCard
-          patient={EXAMPLE_PATIENT}
-          onEditClick={jest.fn()}
-          onDeleteClick={jest.fn()}
-          onExportClick={jest.fn()}
-          onCopyClick={jest.fn()}
-          selected={true}
-          smallScreen={true}
-        />
-      )
+      mantineRecoilWrap(<PatientInfoCard patient={EXAMPLE_PATIENT} {...MOCK_CALLBACK_PROPS} smallScreen={true} />)
     );
     expect(screen.getByLabelText(/menu button/i)).toBeInTheDocument();
   });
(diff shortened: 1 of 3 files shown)


__tests__/components/utils/PatientInfoCard.test.tsx
```diff
@@ -165,4 +166,21 @@ describe('PatientInfoCard', () => {
     const populationSelector = screen.getByPlaceholderText(/select populations/i) as HTMLInputElement;
     expect(populationSelector).toBeInTheDocument();
   });
+
+  it('should render a patient actions menu when the screen is small', () => {
+    render(
+      mantineRecoilWrap(
```
sarahmcdougall, line 172: Any chance we can use the mock callback props here to make this cleaner? I think we might be able to use them and omit `smallScreen`, similar to what we do for the patient in these unit tests

sarahmcdougall (changes requested):

components/utils/PatientInfoCard.tsx
```diff
@@ -43,68 +45,115 @@ export default function PatientInfoCard({
...
-                color="red"
-                variant="subtle"
-              >
-                <Trash />
-              </Button>
-            </Tooltip>
-          </Center>
+        <Grid.Col span={6}>
+          <Group position="right">
+            {!smallScreen ? (
```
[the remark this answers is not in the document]
elsaperelli: true!

__tests__/components/utils/PatientInfoCard.test.tsx
elsaperelli: Ah good call!!

PR #48, opened 2022-09-27
FHIR Utils Refactor

mgramigna:
# Summary

Separates the fhir.ts util file into sub files for individual purposes. Should be clear what they are

## New Behavior

None

## Code Changes

* Move util functions to individual files (content of functions is exactly the same)
* Fixed casing of interfaces in the primary date path stuff (camelCase interfaces bad!!)

# Testing Guidance

* Visual inspection and test all features


sarahmcdougall (changes requested):
Great refactor :) Everything is a lot more organized now. I left a few suggestions for typing, and have two other things:
1. If we want to fix all the interfaces, I noticed the `elementChoice` interface in `parseCodePath.ts` should be changed to `ElementChoice`
2. Do we want to break up the `fhir.test.ts` file at all? Either into separate files for resource creation, date, etc., or putting the fixtures in a separate file? that file is getting pretty big

util/fhir/resourceCreation.ts
```diff
@@ -0,0 +1,152 @@
...
+  return resources;
+}
+
+/**
+ * Creates a copy of the passed in patient object (without references maintained) and updates the
+ * id and identifier as well as creating a new name to differentiate the new patient copy
+ * @param copyPatient {fhir4.Patient} a fhir Patient object to copy
+ * @returns {fhir4.Patient} the new fhir patient copy
+ */
+export function createCopiedPatientResource(copyPatient: fhir4.Patient): fhir4.Patient {
```
elsaperelli, line 68: Even though the name is createCopiedPatientResource, this function creates a copy of a the passed in patient object. Since the return type is fhir4.Patient, should we put this in `patient.ts`? Open to hear the reasoning for keeping this here though!

util/fhir/resourceCreation.ts
```diff
@@ -0,0 +1,152 @@
...
+}
+
+/**
+ * Creates a string representing a patient bundle resource. Creates using a patient resource and
+ * an array of the patient's associated resources
+ * @param {Object} patient FHIR Patient object
+ * @param {Array} resources array of FHIR resources associated with the patient
+ * @returns {String} representation of a FHIR patient bundle resource
+ */
+export function createPatientBundle(patient: fhir4.Patient, resources: fhir4.FhirResource[]): fhir4.Bundle {
```
elsaperelli, line 102: Kind of similar comment as above, would it make more sense for this to be in `patient.ts`? I could be very wrong :)

elsaperelli (changes requested):
I didn't realize how badly we needed this refactor until now! Look awesome. Just two comments about the placement of two functions :)

util/fhir/dates.ts
```diff
@@ -0,0 +1,156 @@
...
+  }
+}
+
+/**
+ * Determines the ideal date info type and path to populate a given attribute with date info
+ * @param {Object} fieldTypeInfo object containing information on date types the specified field can accept
+ * @param {String} path the attribute name that needs to be populated with date info
+ * @returns {Object} an object containing the newPath to enter date info and the ideal date info type
+ */
+export function getDateType(fieldTypeInfo: DateFieldInfo, path: string) {
```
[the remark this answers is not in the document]
mgramigna: should never use `object` as a type in TS, but I can define the actual structure of the return value instead

mgramigna:
>     1. If we want to fix all the interfaces, I noticed the `elementChoice` interface in `parseCodePath.ts` should be changed to `ElementChoice`

Good catch!

>     2. Do we want to break up the `fhir.test.ts` file at all? Either into separate files for resource creation, date, etc., or putting the fixtures in a separate file? that file is getting pretty big

Ah yes definitely. I meant to do this but forgot lol

util/fhir/dates.ts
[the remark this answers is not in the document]
sarahmcdougall: sounds good, that was what vscode suggested so I just went with it lol but makes sense

util/fhir/resourceCreation.ts
mgramigna: I went with the mantra that anything that's generating a FHIR resource at all goes in resourceCreation, even if that thing is a patient. It's definitely not perfect but I think it might be fine? Idk at this point

elsaperelli: Ah I see! That works for me

util/fhir/dates.ts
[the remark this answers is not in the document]
mgramigna: what the heck VS Code! Be better!

elsaperelli (approved):
Looks great to me !

sarahmcdougall (approved):
🔥

PR #49, opened 2022-09-27
Updated TSC config to target es2017 and fqm-execution github dependency

elsaperelli:
# Summary
Updates the TS compiler config to target ES2017, which allows for the use of native async/await as well as changes the fqm-execution dependency to the github link.

## New Behavior
- None

## Code Changes
- `package-lock.json` and `package.json` - changed to fqm-execution github dependency.
- `tsconfig,json` - changed target to ES2017.

# Testing Guidance
- Run `npm run check` to make sure tests still pass and there are no lint or prettier errors or warnings.
- Run `npm run dev` to make sure that everything still works the same.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2847,7 +2847,7 @@
       }
     },
     "cql-execution": {
-      "version": "github:projecttacoma/cql-execution#36346a0fe1526abf48bfe8bdb37fe9e8164292bb",
+      "version": "github:projecttacoma/cql-execution#d1c0e9341c763bb8cc641a5a702cedb0c8385c88",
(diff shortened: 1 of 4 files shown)


sarahmcdougall (approved):
Looks good - I checked the commit hashes for cql-execution and fqm-execution and they're both up to date 👍

mgramigna:
While we're here maybe we can add the elm json caching flag as true to the calculation options in the measure calculation file

hossenlopp (approved):

sarahmcdougall (approved):

PR #51, opened 2022-10-05
Cqfm test case Measure Report

natjoe4:
# Summary
Added functionality for generating cqfm-test-case MeasureReports for each test case to be exported as part of the test case bundle

## New Behavior
when exporting a test case, there should now be an additional MeasureReport resource which is a cqfm-test-case MeasureReport

## Code Changes
- added functionality for creating the test case measure report in resourceCreation.ts
- appropriately updated all places where information that should cause the cqfm-test-case MeasureReport to update
- added testing

# Testing Guidance
- `npm run check`
- `npm run dev`
- upload a measure
- create a patient
- export the patient and ensure that a cqfm-test-case MeasureReport exists in the exported bundle
- import a patient without a cqfm-test-case MeasureReport in its bundle
- export the imported patient and ensure that a cqfm-test-case MeasureReport is present in the bundle
- import multiple patients, then export and ensure that a cqfm-test-case MeasureReport is present in each bundle
- return to the measure upload page and change the measure and measurement period
- return to the reseource creation page and export all resources. Note that the cqfm-test-case MeasureReport will now reflect data from the new measure.
- Link to [cqfm-test-case MeasureReport docs](https://build.fhir.org/ig/HL7/cqf-measures/StructureDefinition-test-case-cqfm.html)
- Ensure all present data are as specified in the docs


__tests__/utils/fhir/resourceCreation.test.ts
```diff
@@ -438,3 +521,45 @@ describe('createCopiedPatientResource', () => {
     expect(patient.id).not.toEqual(PATIENT_RESOURCE.id);
   });
 });
+
+describe('generateTestCaseMRGroup', () => {
+  test.only('returns a group array with all populations from MeasureReport', () => {
```
elsaperelli, line 526:
```suggestion
  it('returns a group array with all populations from MeasureReport', () => {
```
Not a huge deal, but to be consistent with the rest of the file maybe use "it" instead of "test" :)

util/fhir/resourceCreation.ts
```diff
@@ -150,3 +164,89 @@ export function createFHIRResourceString(
...
+  return {
+    resourceType: 'MeasureReport',
+    id: uuidv4(),
+    measure: measure.id as string,
+    period: measurementPeriod,
+    status: 'complete',
+    type: 'individual',
+    meta: {
+      profile: ['http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/test-case-cqfm']
+    },
```
mgramigna, line 193: Wow, above and beyond with the `meta.profile`! love it

state/atoms/cqfmTestMRLookup.ts
```diff
@@ -0,0 +1,6 @@
+import { atom } from 'recoil';
+
+export const cqfmTestMRLookupState = atom<Record<string, fhir4.MeasureReport>>({
```
mgramigna, line 3: Is there any reason this can't be a selector instead? Theoretically it's a computable value that depends on each value in the currentPatients state, the current measure/measurement period, and the desired populations on a test case. 

This should all be computable using a selector, and then you never need to use produce anywhere

__tests__/utils/fhir/resourceCreation.test.ts
```diff
@@ -376,7 +459,7 @@ describe('createPatientBundle', () => {
     expect(createPatientBundle(PATIENT_RESOURCE, [])).toEqual(expectedBundle);
   });
 
-  it('shoult create bundle with patient and resources', () => {
+  it('should create bundle with patient and resources', () => {
```
mgramigna, line 462: thanks

mgramigna (changes requested):
Good stuff. Main comment is about potential use of a selector which would theoretically remove all uses of produce that you added across the app. LMK what you think

state/selectors/cqfmTestMRLookup.ts
```diff
@@ -0,0 +1,29 @@
+import { selector } from 'recoil';
+import { createCQFMTestCaseMeasureReport } from '../../util/fhir/resourceCreation';
+import { measureBundleState } from '../atoms/measureBundle';
+import { measurementPeriodState } from '../atoms/measurementPeriod';
+import { patientTestCaseState } from '../atoms/patientTestCase';
+
+interface CqfmTestMRLookupType {
+  [id: string]: fhir4.MeasureReport;
+}
+export const cqfmTestMRLookupState = selector<CqfmTestMRLookupType>({
```
mgramigna, line 10:
```suggestion
export const cqfmTestMRLookupState = selector<Record<string, fhir4.MeasureReport>>({
```

Nitpick, but this prob makes more sense as a `Record` type

mgramigna: This requires changing the type in the reduce too actually, and deleting the interface

state/selectors/cqfmTestMRLookup.ts
```diff
@@ -0,0 +1,29 @@
...
+  key: 'cqfmTestMRLookupState',
+  get: ({ get }) => {
+    const measureBundle = get(measureBundleState);
+    const { start, end } = get(measurementPeriodState);
+    const currentPatients = get(patientTestCaseState);
+    if (measureBundle.content && start && end) {
+      return Object.keys(currentPatients).reduce((acc: CqfmTestMRLookupType, e) => {
```
mgramigna, line 17: I think the usage of reduce does make sense given that we're assembling an object from an array, can we at least be a little clearer about the variable names though? It can get a little confusing especially if you don't know how reduce works. Suggestions:

`acc` -> `lookupResult`
`e` -> `patientId`

util/fhir/resourceCreation.ts
```diff
@@ -150,3 +164,94 @@ export function createFHIRResourceString(
...
+    extension: [
+      {
+        url: 'http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/cqfm-inputParameters',
+        valueReference: {
+          reference: `#${parametersId}`
+        }
+      }
+    ],
+
```
mgramigna, line 203:
```suggestion
```

components/patient-creation/PatientCreationPanel.tsx
```diff
@@ -303,6 +305,7 @@ function PatientCreationPanel() {
             }
 
             draftState[testCase.patient.id] = testCase;
+            addedIds.push(testCase.patient.id);
```
mgramigna, line 308: Is this variable actually used for anything?

mgramigna (changes requested):
Few more small changes/nitpicks

mgramigna (approved):

elsaperelli (approved):
Nice!

PR #53, opened 2022-10-12
Population comparison table

elsaperelli:
# Summary
This PR adds a table to the `MeasureHighlightingPanel` that displays a comparison of the desired populations and actual populations that the patient falls under. 

## New Behavior
When a patient is selected, a population comparison table is rendered in the `MeasureHighlightingPanel` above the highlighted logic. This table has two columns of values: Desired and Actual. The rows are the populations that exist on the uploaded measure bundle. The values in the Desired row represent whether or not that population is included in the Desired population (1 for included, 0 for not). The values in the Actual row represent whether or not that population exists on the measure calculation measure report (1 for exists, 0 for does not). Additionally, when rows of the table match, they are highlighted green. When the rows do not match, they are highlighted red.

## Code Changes
- `components/utils/PopulationMultiSelect.tsx` - moved the `getMeasurePopulations` function to a new file (`MeasurePopulations.tsx`) so it can get reused.
- `components/calculation/MeasurePopulations.tsx` - added the `getMeasurePopulations` function.
- `components/calculation/MeasureHighlightingPanel.tsx` - added functions in order to populate the population comparison table, renders the population comparison table.
- `__tests__/components/calculation/MeasureHighlightingPanel.test.tsx` - added unit test for the population comparison table.

# Testing Guidance
1. Run `npm run check` to ensure that all unit tests pass and there are no lint/prettier errors or warnings.
2. Upload a measure bundle and create or upload a patient (make sure to do both at some point).
3. Select the patient without any desired populations added and make sure that the desired population column is populated with all 0s.
4. Add desired populations and ensure that the correct values in the table are changing. 
5. Add measure resources to get the patient in any of the populations and ensure that those values change in the table as well.
7. I am purposely leaving this kind of open ended so that any bugs that I am not thinking of can be caught, but let me know if you have any questions or need help!

### DIFF
--- a/components/calculation/MeasureHighlightingPanel.tsx
+++ b/components/calculation/MeasureHighlightingPanel.tsx
@@ -1,13 +1,8 @@
-import { ActionIcon, createStyles, Group, Popover, Table, Text } from '@mantine/core';
+import { createStyles } from '@mantine/core';
 import { useRecoilValue } from 'recoil';
 import parse from 'html-react-parser';
 import { measureReportLookupState } from '../../state/atoms/measureReportLookup';
-import { patientTestCaseState } from '../../state/atoms/patientTestCase';
-import { measureBundleState } from '../../state/atoms/measureBundle';
-import { useMemo, useState } from 'react';
-import { getMeasurePopulations, MultiSelectData } from './MeasurePopulations';
-import { MeasureReportGroup } from 'fhir/r4';
-import { InfoCircle } from 'tabler-icons-react';
+import PopulationComparisonTable from './PopulationComparisonTable';
 
 const useStyles = createStyles({
   highlightedMarkup: {
@@ -18,18 +13,6 @@ const useStyles = createStyles({
   panel: {
     maxHeight: '100%',
     overflow: 'scroll'
-  },
-  highlightRed: {
-    backgroundColor: '#edd8d0',
-    color: '#a63b12',
-    borderBottomColor: '#a63b12',
-    borderBottomStyle: 'double'
-  },
-  highlightGreen: {
-    backgroundColor: '#ccebe0',
-    color: '#20744c',
-    borderBottomColor: '#20744c',
-    borderBottomStyle: 'solid'
   }
 });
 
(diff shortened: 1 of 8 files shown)


components/calculation/MeasureHighlightingPanel.tsx
```diff
@@ -12,20 +18,146 @@ const useStyles = createStyles({
...
+              <Popover.Target>
+                <ActionIcon aria-label={'More Information'} onClick={() => setOpened(o => !o)}>
+                  <InfoCircle size={20} />
+                </ActionIcon>
+              </Popover.Target>
+              <Popover.Dropdown>
+                The rows in this table are the populations and the columns two columns represent the desired and actual
+                populations. The contents are either 0 or 1: 0 if that population does not exist, 1 if it does. If any
+                of the rows are highlighted red, that means the desired population does not match the actual population.
+                If they are highlighted green, then they match.
```
natjoe4, line 122: I can't suggest a multiline comment here, so I'll just add some edits in a regular comment :)

```
The rows in this table represent the possible populations as specified on the uploaded Measure and the two columns represent the desired and actual populations as stored/calculated by FQM-Testify. For the 'Actual' column, the value will be 1 if the patient calculates into the respective population, and 0 if it does not. For the 'Desired' column, the value will be 1 if the respective population was selected as a desired population for the patient, and 0 if not. If any of the rows are highlighted red, that means the desired population does not match the actual population.
If they are highlighted green, then they match.
```

components/calculation/MeasureHighlightingPanel.tsx
```diff
@@ -12,20 +18,146 @@ const useStyles = createStyles({
...
+  }
+
+  if (measure) {
+    const group = measureReportLookup[patientId]?.group?.[0];
+    const bothPopulations = constructBothPopulationsValuesArray(group);
+
+    return (
+      <>
+        <div />
+        <Group>
```
natjoe4, line 105: I think this table is big enough to be separated into its own component

components/calculation/MeasureHighlightingPanel.tsx
```diff
@@ -12,20 +18,146 @@ const useStyles = createStyles({
...
+   */
+  function constructBothPopulationsValuesArray(group: MeasureReportGroup | undefined) {
+    const desiredPopulations = constructDesiredPopulationsValuesArray(getMeasurePopulations(measure));
+    const bothPopulations: BothPopulations = {};
+    group?.population?.map(population => {
+      const key = population?.code?.coding?.[0]?.display || population?.code?.coding?.[0]?.code;
+      if (key) {
+        bothPopulations[key as string] = { desired: desiredPopulations[key as string], actual: population?.count };
+      }
+    });
```
natjoe4, line 94: A little unconventional to have a `.map` here with no return value. I would swap this to a forEach, or if you're feeling really fancy, go for a reduce. This comment also applies to the above function

components/calculation/MeasureHighlightingPanel.tsx
```diff
@@ -12,20 +18,146 @@ const useStyles = createStyles({
...
+    backgroundColor: '#edd8d0',
+    color: '#a63b12',
+    borderBottomColor: '#a63b12',
+    borderBottomStyle: 'double'
+  },
+  highlightGreen: {
+    backgroundColor: '#ccebe0',
+    color: '#20744c',
+    borderBottomColor: '#20744c',
+    borderBottomStyle: 'solid'
```
natjoe4, line 32: For these colors, we probably want to use the pre-build Mantine theme colors to give the app a uniform look. You can pass in `theme` as an argument to the createStyles function, then access `theme.colors.red[x]` or `theme.colors.green[x]`, where x is a number from 1-10 determining the darkness of the color, which will supply an array of colors available in the app.

natjoe4 (changes requested):
Looks great! I did run into a bug where if you edit the patient after selecting desired populations, the desired populations column on the table resets to all zeros. Otherwise just small comments. Nice work!

components/calculation/MeasureHighlightingPanel.tsx
natjoe4: Ahh, I see that these were selected to match the measure highlighting colors, which we don't have access to change in testify, so disregard this comment :)

components/calculation/MeasureHighlightingPanel.tsx
elsaperelli: This is a good point

natjoe4 (approved):
Awesome work! Tested rigorously with created and imported patients

components/calculation/MeasurePopulations.tsx
```diff
@@ -0,0 +1,45 @@
+import { Enums } from 'fqm-execution';
```
hossenlopp, line 1: This file isn't really a component and should probably be just included in the PopulationComparisonTable file or moved to the `utils` folder.

components/calculation/MeasurePopulations.tsx
```diff
@@ -0,0 +1,45 @@
...
+}
+
+/**
+ * Compiles an array of data for the MultiSelect component, using population codes
+ * from the measure resource, across all groups in the measure.
+ *
+ * Excludes measure-population, measure-population-exclusion, and measure-observation
+ * (used for CV measures).
+ * @param {fhir4.Measure} measure - FHIR measure resource
+ * @returns {Array} array of strings representing all unique measure populations
```
hossenlopp, line 16: This isn't returning an array of strings, update this comment.

hossenlopp (changes requested):
Functions well after fixes from Nat's comments. Just a few comments on the shared function you moved.

natjoe4 (approved):
🎉

hossenlopp (approved):

PR #56, opened 2022-11-22
Made header flush with the test case editor

elsaperelli:
# Summary
Makes the test case editor panels flush against the header.

## New Behavior
No new behavior.

## Code Changes
- `components/TestCaseEditor.tsx` - Changes the height of the grid from `85vh` to be the height of the window subtracted by the height of the header so that it fits the full screen.
- `pages/_app.tsx` - Changes the margins of the `AppShell` because Mantine automatically gives them 16px of padding.
- `pages/generate-test-cases.tsx` - Removed the top padding.
- `pages/index.tsx` - Delete a `<Head>` that we no longer need because it is in `_app.tsx`.

# Testing Guidance
- Run `npm run check` to ensure that there are no lint errors or warnings and that all tests still pass.
- `npm run dev` and make sure that the UI of the app is consistent in different window sizes and looks good overall. 
- Let me know if you have any comments or suggestions!

Before:
image

After: 
image

### DIFF
--- a/components/TestCaseEditor.tsx
+++ b/components/TestCaseEditor.tsx
@@ -64,7 +64,7 @@ export default function TestCaseEditor() {
 
   return (
     <>
-      <Grid style={{ height: '85vh' }}>
+      <Grid style={{ height: 'calc(100vh - 120px)' }}>
         <Grid.Col span={3} className={classes.panel}>
           <PatientCreationPanel />
         </Grid.Col>
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -15,7 +15,7 @@ export default function App(props: AppProps) {
   return (
     <RecoilRoot>
       <Head>
-        <title>FQM Testify: an eCQM Analysis Tool</title>
+        <title>FQM Testify: a FHIR-based eCQM Analysis Tool</title>
         <meta name="viewport" content="minimum-scale=1, initial-scale=1, width=device-width" />
         <link rel="icon" href="/favicon.ico" />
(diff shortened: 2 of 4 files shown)


pages/_app.tsx
```diff
@@ -15,14 +15,17 @@ export default function App(props: AppProps) {
   return (
     <RecoilRoot>
       <Head>
-        <title>FQM Testify: an ECQM Analysis Tool</title>
+        <title>FQM Testify: an eCQM Analysis Tool</title>
```
sarahmcdougall, line 18:
```suggestion
        <title>FQM Testify: a FHIR-based eCQM Analysis Tool</title>
```
Might be good to add this so that it aligns with the README

sarahmcdougall (approved):
Just left one optional suggestion but otherwise looks good to me! Tested on both Chrome and Safari with different window sizes and it looked good

pages/_app.tsx
elsaperelli: Nice!

components/TestCaseEditor.tsx
```diff
@@ -64,7 +64,7 @@ export default function TestCaseEditor() {
 
   return (
     <>
-      <Grid style={{ height: '85vh' }}>
+      <Grid style={{ height: 'calc(100vh - 120px)' }}>
```
natjoe4, line 67: Didn't know you could do this! Good find :)

natjoe4 (approved):
Great work. Looks much better 🎉

PR #57, opened 2022-11-28
Test case Measure Report Import

natjoe4:
# Summary
Importing a patient bundle with a cqfm test case MeasureReport will now update that patient's desired population.

## New Behavior
When importing a patient bundle or a .zip file of patient bundles with cqfm test case Measure Report resources, the desired population of the patient test case will reflect the populations scored in the cqfm test case measure report. cqfm test case MeasureReports will no longer be viewable in the app when imported as part of a patient bundle.

## Code Changes
 - Updated bundleToTestCase function to take new `populationGroups` parameter, detect presence of test case MeasureReport, parse it into desired populations when appropriate, and exclude it from resource creation.
 - Created new isTestCaseMeasureReport helper function
 - Converted existing `getMeasurePopulations` function to `getMeasurePopulationsForSelection`
 - Created new `getMeasurePopulations` function that returns just the codes from the populations

# Testing Guidance
- `npm run check`
- `npm run dev`
- upload `EXM130` bundle
- Try uploading the test cases from the attached file. Unzip the first .zip file to reveal individual test cases and a nested .zip file. Uploading the initial .zip as is will throw an error
- The test cases are named after the expected outcome for each upload (ex. Nancy McNumerator desires to be in the numerator)
- `Wannabe Numex` and `Multiple Test Case Mrs` should both throw comprehensive errors on attempted upload
[test-cases.zip](https://github.com/projecttacoma/fqm-testify/files/10107666/test-cases.zip)


util/import.ts
```diff
@@ -1,11 +1,50 @@
...
+  const testCaseMeasureReportArr = bundle.entry.filter(isTestCaseMeasureReport);
+  if (testCaseMeasureReportArr.length > 1) {
+    // TODO: Once we have import errors persist on page, replace this!!!
+    throw new Error(
+      `Expected 1 or 0 test case measure reports in bundle, but found ${testCaseMeasureReportArr.length}`
+    );
+  }
+  if (testCaseMeasureReportArr.length === 1) {
+    // use reduce here to retrieve population codes and filter out potential undefined codes
+    let retrievedPops = (testCaseMeasureReportArr[0].resource as fhir4.MeasureReport)?.group?.[0]?.population?.reduce(
```
sarahmcdougall, line 19: Something that probably warrants a larger discussion is how to handle having multiple groups. For the most part, I think we just look at the first group in the array, but with the advancement of ratio measure support and other issues brought up by SB, I think we should make sure we are consistent with whether we look at the first group versus loop over all groups.

Here, for example, we look at the first group, but in `getMeasurePopulations()` (from `util/MeasurePopulations.ts`) we loop over all the groups. Should we change the logic in `getMeasurePopulations()` to only look at the first group since we are only looking at the first group here?

util/MeasurePopulations.ts
```diff
@@ -43,3 +43,22 @@ export function getMeasurePopulations(measure: fhir4.Measure): MultiSelectData[]
   });
   return measurePopulations;
 }
+
+/**
+ * Extracts all population codes included in measure groups
```
sarahmcdougall, line 48: Based on Matt's [typescript guide](https://gist.github.com/mgramigna/bdc933fae2150216681c28d1f0f7aac7) (shoutout @mgramigna the king 👑 of typescript), I don't think we need to include the JSDoc param/return since it's not adding a lot of new information and can easily become outdated if we update this function in the future

sarahmcdougall (changes requested):
Functionality looks good when testing and the test cases in the .zip align nicely with the FHIR spec in regard to the attributes that should be included in a test case measure report.

Open question: Should we add more context about the test case measure report for the user? I'm thinking of a case where the user does not realize the uploaded test case includes a test case measure report. In this case, it might take some time for the user to understand where the pre-populated desired populations came from. I'm not really sure how often this would come up (maybe we can talk about this more with the TA team to get more insight). I was thinking that, minimally, we could add more text to the "Desired Populations" info button that says something along the lines of "Pre-selected desired populations come from the imported test case measure report, if present" + a link to the spec. Otherwise, it doesn't seem like there's a way for the user to access the measure report from the app to easily see what's included in the measure report.

lmd59:
> Functionality looks good when testing and the test cases in the .zip align nicely with the FHIR spec in regard to the attributes that should be included in a test case measure report.
> 
> Open question: Should we add more context about the test case measure report for the user? I'm thinking of a case where the user does not realize the uploaded test case includes a test case measure report. In this case, it might take some time for the user to understand where the pre-populated desired populations came from. I'm not really sure how often this would come up (maybe we can talk about this more with the TA team to get more insight). I was thinking that, minimally, we could add more text to the "Desired Populations" info button that says something along the lines of "Pre-selected desired populations come from the imported test case measure report, if present" + a link to the spec. Otherwise, it doesn't seem like there's a way for the user to access the measure report from the app to easily see what's included in the measure report.

We could also or alternatively put something in the import results message to show that a test case patient and test case measure report has been imported. This would at least be informative of what's contained in the upload if the user is unaware.

util/import.ts
```diff
@@ -1,11 +1,50 @@
 import { TestCaseInfo } from '../state/atoms/patientTestCase';
 
-export function bundleToTestCase(bundle: fhir4.Bundle): TestCaseInfo {
+export function bundleToTestCase(bundle: fhir4.Bundle, populationGroupCodes: string[]): TestCaseInfo {
```
lmd59, line 3: Would recommend a function comment on this, especially since it's grown in scope.

lmd59 (commented):
Function looks good and code looks good! Small comment and suggestion.

util/import.ts
```diff
@@ -1,11 +1,56 @@
 import { TestCaseInfo } from '../state/atoms/patientTestCase';
 
-export function bundleToTestCase(bundle: fhir4.Bundle): TestCaseInfo {
+/**
```
lmd59, line 3: I feel like I'm being a bit persnickety, but would you mind adding the jsdoc style @param and @returns bits? A part of my motivation was that when I first looked at it, I couldn't tell whether populationGroupCodes was the codes that were desired for the test case or the codes that were relevant to the measure. I figured it out, but I'm inclined toward making it as easy as possible. I know that style guide suggests "Avoid uses of typed JSDoc parameters and returns when possible", but I think this is a case where it's beneficial (and chatted with Matt about where it makes sense to draw that line, so style guide may get a couple of bullets updated for that section).

lmd59 (approved):
Awesome- thanks!!

sarahmcdougall (approved):
Changes look good :) Good call adding TODO's for the multiple group issues so that it's easier to find and make the appropriate changes later

PR #62, opened 2023-04-04
Mrs input

natjoe4:
# Summary
FQM-Testify can now import a Measure from a passed in Measure Repository Service URL

## New Behavior
All old behavior should be unchanged!
There is now a new option on the Measure Upload page to upload from Measure Repository Service
The app should work identically once a valid Measure is uploaded through this new method.

## Code Changes

- Created MeasureRepositoryUpload.tsx, a component that pulls measures from a MRS and then packages the selected measure with all dependencies into a Measure Bundle to pass into testify
- Created MeasureRepositoryHeader.tsx, a component that frames the above component and adds helpful information
- Created displayMapToSelectData Selector to populate Measure Select component.
- Added additional information to MeasureBundleState to determine the type of upload and persist information for both file and MRS sourced upload when leaving/returning to the page
- Minor updates to other components that access MeasureBundleState to accommodate these changes
- Minor updates to existing tests to accommodate changes

# Testing Guidance
- `npm run check`
- Run through the regular file upload workflow and ensure nothing has changed
- Start up our Measure Repository Service and upload as many valid and invalid bundles as you'd like
- Start up Testify with `npm run dev`
- Type in the url of the running measure repository service and ensure the call is successful and the measure select dropdown appears and is populated accurately
- Select a working measure and ensure that Testify workflow is unchanged from there
- Select a broken measure and ensure that the error handling is intuitive and properly labeled
- Try to break the system. Move back and forth between screens. Upload a file when a bundle is already uploaded through MRS and vice versa. Change the url after uploading a bundle. Make sure the code works intuitively throughout the process.

### DIFF
--- a/components/measure-upload/MeasureRepositoryUpload.tsx
+++ b/components/measure-upload/MeasureRepositoryUpload.tsx
@@ -58,7 +58,7 @@ export default function MeasureRepositoryUpload({ logError }: MeasureUploadProps
 
   /**
    * Handles updating of the Measure Repository Service url field and associated state variables
-  */
+   */
   function handleMrsUrlChange(event: React.ChangeEvent<HTMLInputElement>) {
     if (measureSelectIsOpen) {
       setMeasureSelectIsOpen(false);
@@ -76,7 +76,7 @@ export default function MeasureRepositoryUpload({ logError }: MeasureUploadProps
   /**
    * Takes the entries from a Measure retrieval response and creates a map from Measure id
    * to Measure display label for Measure select component
-  */
+   */
   function createMeasureDisplayMap(
     displayMap: Record<string, string>,
     { resource: measure }: { resource: fhir4.Measure }
@@ -100,7 +100,7 @@ export default function MeasureRepositoryUpload({ logError }: MeasureUploadProps
   /**
    * Sends a `GET` request to the `/Measure` endpoint of the specified Measure Repository Service.
    * Parses and stores the measureIds if successful
-  */
+   */
   async function retrieveMeasures() {
     setIsLoadingIds(true);
     let parsedMeasureRepositoryUrl = measureRepositoryUrl;
@@ -150,7 +150,7 @@ export default function MeasureRepositoryUpload({ logError }: MeasureUploadProps
   /**
    * Sends a `POST` request to the `Measure/:id/$package` endpoint of the specified Measure
    * Repository Service. Checks the resulting bundle is valid and saves it if so.
-  */
+   */
   async function retrieveMeasurePackage(id: string) {
     setMeasureBundle(mb => ({ ...mb, selectedMeasureId: id }));
     if (id) {
--- a/components/measure-upload/MeasureRepositoryUpload.tsx
+++ b/components/measure-upload/MeasureRepositoryUpload.tsx
@@ -1,13 +1,22 @@
-import { Button, Collapse, Group, Select, TextInput } from '@mantine/core';
+import { Button, Collapse, Grid, Group, Loader, Select, TextInput } from '@mantine/core';
 import React, { useState } from 'react';
 import { useEffect } from 'react';
-import { useRecoilState, useSetRecoilState } from 'recoil';
+import { useRecoilState, useRecoilValue, useSetRecoilState } from 'recoil';
 import { measureBundleState } from '../../state/atoms/measureBundle';
 import { measurementPeriodState } from '../../state/atoms/measurementPeriod';
 import { identifyMissingValueSets, MeasureUploadProps, populateMeasurementPeriod } from '../../util/MeasureUploadUtils';
 import { showNotification } from '@mantine/notifications';
 import { v4 as uuidv4 } from 'uuid';
 import { IconAlertCircle } from '@tabler/icons';
+import { CircleCheck } from 'tabler-icons-react';
+import { displayMapToSelectDataState } from '../../state/selectors/displayMapToSelectData';
+
+enum PACKAGE_STATES {
+  LOADING,
+  SUCCESS,
+  FAIL,
+  NONE
+}
 
 const MEASURE_RETRIEVAL_BODY = {
   resourceType: 'Parameters',
@@ -26,21 +35,22 @@ const MEASURE_RETRIEVAL_OPTIONS = {
 };
 
 export default function MeasureRepositoryUpload({ logError }: MeasureUploadProps) {
-  const [mrsUrl, setMrsUrl] = useState('');
-  const [measureIds, setMeasureIds] = useState([]);
-  const [selectedId, setSelectedId] = useState<string | null>(null);
-
-  const [measureBundle, setMeasureBundle] = useRecoilState(measureBundleState);
-  const setMeasurementPeriod = useSetRecoilState(measurementPeriodState);
-
-  const [isLoading, setIsLoading] = useState(false);
+  const [isLoadingIds, setIsLoadingIds] = useState(false);
+  const [isLoadingPackage, setIsLoadingPackage] = useState<PACKAGE_STATES>(PACKAGE_STATES.NONE);
   const [measureSelectIsOpen, setMeasureSelectIsOpen] = useState(false);
 
+  const [{ measureRepositoryUrl, isFile, selectedMeasureId, displayMap }, setMeasureBundle] =
+    useRecoilState(measureBundleState);
+  const setMeasurementPeriod = useSetRecoilState(measurementPeriodState);
+
+  const idSelectData = useRecoilValue(displayMapToSelectDataState);
+
   useEffect(() => {
-    if (mrsUrl && selectedId) {
-      retrieveMeasurePackage();
+    if (isFile) {
+      setMeasureSelectIsOpen(false);
+      setIsLoadingPackage(PACKAGE_STATES.NONE);
     }
-  }, [selectedId]);
+  }, [isFile]);
 
   const rejectUpload = (
     message: string | string[],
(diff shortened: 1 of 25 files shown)


components/measure-upload/MeasureRepositoryUpload.tsx
```diff
@@ -0,0 +1,235 @@
...
+      id: uuidv4(),
+      message,
+      timestamp: new Date().toISOString(),
+      attemptedBundleDisplay,
+      isValueSetMissingError
+    });
+    showNotification({
+      icon: <IconAlertCircle />,
+      title: 'Bundle upload failed',
+      message: 'There was an issue uploading your bundle. Please correct the issues listed below.',
```
sarahmcdougall, line 70: The error message for measure _file_ upload is: "There was an issue with your file." Should we change the error message in `MeasureFileUpload.tsx` to match this one (since technically you are uploading a bundle there too...), or do you think we should keep the two messages separate to better differentiate what measure upload method caused the error? If you think we should keep two different messages, maybe we can add more context to this one and say something like "There was an issue uploading your bundle _from the server_"

components/measure-upload/MeasureRepositoryUpload.tsx
```diff
@@ -0,0 +1,235 @@
...
+      color: 'red'
+    });
+    setIsLoadingPackage(PACKAGE_STATES.FAIL);
+  };
+
+  /*
+    Handles updating of the Measure Repository Service url field and associated state variables
+  */
```
elsaperelli, line 78:
```suggestion
  /**
   * Handles updating of the Measure Repository Service url field and associated state variables
  */
```
Not sure if this lines up but I think this is what we have in the rest of this repo

components/measure-upload/MeasureRepositoryUpload.tsx
```diff
@@ -0,0 +1,235 @@
...
+      displayMap: {},
+      ...(!mb.isFile ? { content: null } : undefined)
+    }));
+    setIsLoadingPackage(PACKAGE_STATES.NONE);
+  }
+
+  /*
+    Takes the entries from a Measure retrieval response and creates a map from Measure id
+    to Measure display label for Measure select component
+  */
```
elsaperelli, line 96:
```suggestion
  /**
   * Takes the entries from a Measure retrieval response and creates a map from Measure id
   * to Measure display label for Measure select component
  */
```

components/measure-upload/MeasureRepositoryUpload.tsx
```diff
@@ -0,0 +1,235 @@
...
+      label = measure.id as string;
+    }
+    displayMap[measure.id as string] = label;
+    return displayMap;
+  }
+
+  /*
+    Sends a `GET` request to the `/Measure` endpoint of the specified Measure Repository Service.
+    Parses and stores the measureIds if successful
+  */
```
elsaperelli, line 120:
```suggestion
  /**
   * Sends a `GET` request to the `/Measure` endpoint of the specified Measure Repository Service.
   * Parses and stores the measureIds if successful
  */
```

components/measure-upload/MeasureRepositoryUpload.tsx
```diff
@@ -0,0 +1,235 @@
...
+        color: 'red'
+      });
+    }
+    setIsLoadingIds(false);
+  }
+
+  /*
+    Sends a `POST` request to the `Measure/:id/$package` endpoint of the specified Measure
+    Repository Service. Checks the resulting bundle is valid and saves it if so.
+  */
```
elsaperelli, line 154:
```suggestion
  /**
   * Sends a `POST` request to the `Measure/:id/$package` endpoint of the specified Measure
   * Repository Service. Checks the resulting bundle is valid and saves it if so.
  */
```

components/measure-upload/MeasureRepositoryUpload.tsx
```diff
@@ -0,0 +1,235 @@
...
+          title: 'Failed reaching out to server',
+          message: `Server responded with code: ${response.status}:${response.statusText}`,
+          color: 'red'
+        });
+      }
+    } catch (e) {
+      showNotification({
+        icon: <IconAlertCircle />,
+        title: 'Failed reaching out to server',
+        message: `Could not connect to server with url: ${measureRepositoryUrl}`,
```
sarahmcdougall, line 144: As we have it right now, we can click on "Get Measures" without providing any url input, and this error will be thrown and say `"Could not connect to server with url: "` since `measureRepositoryUrl` is an empty string. Instead of this, should we add functionality that will instead gray out the "Get Measures" button when the search bar is empty? That way we will never have this error thrown with an empty `measureRepositoryUrl`

Could also add this as a follow-up task to the backlog if it would cause too much overhead... still getting back into the frontend mindset so not sure how difficult this would be to implement, given the current functionality.

components/measure-upload/MeasureRepositoryUpload.tsx
```diff
@@ -0,0 +1,235 @@
...
+  }
+
+  /*
+    Sends a `GET` request to the `/Measure` endpoint of the specified Measure Repository Service.
+    Parses and stores the measureIds if successful
+  */
+  async function retrieveMeasures() {
+    setIsLoadingIds(true);
+    try {
+      const response = await fetch(`${measureRepositoryUrl}/Measure`);
```
sarahmcdougall, line 124: Nitpick here so I'm open to others' thoughts on this... is it worth adding a check to remove a trailing slash from `measureRepositoryUrl` before doing the `fetch`? I don't think we've added this handling in the past to our FHIR apps but I've noticed that a lot of SMART applications check for trailing slash and it makes things a little easier for the user (the "user" being me... I always type in `http://localhost:3000/4_0_1/` by default lol. not sure if others do this as well)

components/measure-upload/MeasureRepositoryUpload.tsx
```diff
@@ -0,0 +1,235 @@
...
+  const idSelectData = useRecoilValue(displayMapToSelectDataState);
+
+  useEffect(() => {
+    if (isFile) {
+      setMeasureSelectIsOpen(false);
+      setIsLoadingPackage(PACKAGE_STATES.NONE);
+    }
+  }, [isFile]);
+
+  const rejectUpload = (
```
elsaperelli, line 55: Would it be possible to make this a bit more reusable between `MeasureRespositoryUpload.tsx` and `MeasureFileUpload.tsx` and put it in `measureUploadUtils.ts`?

components/measure-upload/MeasureRepositoryUpload.tsx
```diff
@@ -0,0 +1,235 @@
...
+    }
+    displayMap[measure.id as string] = label;
+    return displayMap;
+  }
+
+  /*
+    Sends a `GET` request to the `/Measure` endpoint of the specified Measure Repository Service.
+    Parses and stores the measureIds if successful
+  */
+  async function retrieveMeasures() {
```
elsaperelli, line 121: When a mrs doesn't have any measures, the measure dropdown is disabled which is probably intuitive enough for the user but maybe we could add an alert that says that there are no measures in the provided measure repository service url?

components/measure-upload/MeasureRepositoryUpload.tsx
```diff
@@ -0,0 +1,235 @@
...
+    Sends a `POST` request to the `Measure/:id/$package` endpoint of the specified Measure
+    Repository Service. Checks the resulting bundle is valid and saves it if so.
+  */
+  async function retrieveMeasurePackage(id: string) {
+    setMeasureBundle(mb => ({ ...mb, selectedMeasureId: id }));
+    if (id) {
+      const display = displayMap[id];
+      setIsLoadingPackage(PACKAGE_STATES.LOADING);
+      const response = await fetch(`${measureRepositoryUrl}/Measure/${id}/$package`, MEASURE_RETRIEVAL_OPTIONS);
+      const responseBody = await response.json();
```
sarahmcdougall, line 161: Looks like `responseBody` is showing up as an `any` type... I think we should amp up the typing in this function - since `responseBody` is of type `any`, everything else seems to be as well (like `measures`, etc.). Then we can avoid having to set `measures` as a `fhir4.BundleEntry[]` way later (I think)

components/measure-upload/MeasureRepositoryUpload.tsx
```diff
@@ -0,0 +1,235 @@
...
+      if (measures.length !== 1) {
+        rejectUpload(`Expected exactly 1 Measure in uploaded Bundle, received: ${measures.length}`, display);
+        return;
+      }
+
+      const missingValueSets = await identifyMissingValueSets(responseBody);
+      if (missingValueSets.length > 0) {
+        rejectUpload(missingValueSets, display, true);
+        return;
+      }
```
elsaperelli, line 184: If `rejectUpload` and `rejectFile` can be generalized to be the same function, then this could also be I think.

elsaperelli (changes requested):
Looks great! Just some nit-picky inline comments and suggestions :)

components/measure-upload/MeasureRepositoryUpload.tsx
```diff
@@ -0,0 +1,235 @@
...
+              value={selectedMeasureId}
+              onChange={retrieveMeasurePackage}
+              placeholder="Select ID"
+              searchable
+              clearable
+              nothingFound={'No options matching search'}
+            />
+          </Grid.Col>
+          <Grid.Col span={1} style={{ display: 'flex', alignItems: 'end', justifyContent: 'center' }}>
+            {isLoadingPackage === PACKAGE_STATES.LOADING && <Loader />}
```
sarahmcdougall, line 227: Not sure how much overhead this would be, but I think it would be helpful to add some text alongside the loader saying that the measure is being packaged. When I had no ValueSets cached, the loader was spinning for a while, and so it can be unclear to the user that something is actually occurring under the hood. Then, when successful, we could add some text alongside the `CircleCheck` that says something along the lines of "Measure successfully packaged"

components/measure-upload/MeasureRepositoryUpload.tsx
sarahmcdougall: I like that idea - was going to say the same thing

sarahmcdougall (changes requested):
Amazing work on this - kudos to you for stepping up and working on fqm-testify after not working on it for a while!

Not too many change requests from me - code looks pretty good overall. I left a few suggestions/enhancements to make the UI more friendly. I'll leave it to you to decide what's in-scope/out-of-scope for this task - let me know if there's anything you want to discuss further!

components/measure-upload/MeasureRepositoryUpload.tsx
natjoe4: Hmm I'm leaning towards it being ok to have the same error message here, which is simplest when consolidating the two `rejectUpload/File` functions

components/utils/MeasureRespositoryUploadHeader.tsx
```diff
@@ -0,0 +1,39 @@
...
+              <InfoCircle size={20} />
+            </ActionIcon>
+          </Popover.Target>
+          <Popover.Dropdown style={{ width: '300px' }}>
+            <Text>Enter the base URL for a Measure Repository Service server.</Text>
+            <Text>
+              The server must support the{' '}
+              <Anchor href="https://build.fhir.org/ig/HL7/cqf-measures/measure-repository-service.html#package">
+                $package
+              </Anchor>{' '}
```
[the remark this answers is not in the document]
natjoe4: I think we actually need these :/ otherwise the $package is smooshed into the words on either side

components/measure-upload/MeasureRepositoryUpload.tsx
natjoe4: Hahaha good call. Seems like it'll save some grief, and easy enough to implement

components/utils/UploadErrorInfo.tsx
```diff
@@ -13,7 +13,7 @@ export default function UploadErrorInfo({ error }: UploadErrorInfoProps) {
   return (
     <Paper withBorder shadow="md" p="md">
       <Group spacing="xs">
-        <Text weight={600}>{error.attemptedFileName}</Text>
+        <Text weight={600}>{error.attemptedBundleDisplay}</Text>
         <Text color="dimmed">{error.timestamp}</Text>
       </Group>
       {error.isValueSetMissingError ? (
```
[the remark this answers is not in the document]
natjoe4: I added text to the error when a vsError is thrown from the measure repository section. Do you think that's sufficient?

components/utils/MeasureRespositoryUploadHeader.tsx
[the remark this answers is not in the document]
elsaperelli: LOL I took them out of the wrong place when I tested this my bad

elsaperelli (approved):
Nice! Lgtm

components/measure-upload/MeasureRepositoryUpload.tsx
```diff
@@ -120,10 +103,26 @@ export default function MeasureRepositoryUpload({ logError }: MeasureUploadProps
   */
   async function retrieveMeasures() {
     setIsLoadingIds(true);
+    let parsedMeasureRepositoryUrl = measureRepositoryUrl;
+    if (measureRepositoryUrl.slice(-1) === '/') {
```
sarahmcdougall, line 107: If I enter a measure repository url and then delete it, the "Get Measures" button is grayed out. However, for me the "Get Measures" button is not grayed out when I first load or refresh the app. If I click on "Get Measures" without a url, I get a TypeError because it's trying to slice a nonexistent url:

Screenshot 2023-04-06 at 1 34 01 PM

We should make the button grayed out on initial load if possible, or at least do a check here to see that the url is defined

components/measure-upload/MeasureRepositoryUpload.tsx
```diff
@@ -156,39 +155,62 @@ export default function MeasureRepositoryUpload({ logError }: MeasureUploadProps
     setMeasureBundle(mb => ({ ...mb, selectedMeasureId: id }));
     if (id) {
       const display = displayMap[id];
+      showNotification({
+        loading: true,
+        title: `Packaging ${display}...`,
+        message: `Waiting for measure repository service to finish packaging Measure: ${display}`,
```
sarahmcdougall, line 161: I think this is a great addition! Makes it easy to see what's going on

components/utils/UploadErrorInfo.tsx
```diff
@@ -24,8 +24,16 @@ export default function UploadErrorInfo({ error }: UploadErrorInfoProps) {
               justifyContent: 'space-between'
             }}
           >
-            <Text>Bundle is missing required ValueSets</Text>
-            <Button variant="subtle" color="gray" onClick={() => setIsDetailOpen(!isDetailOpen)} style={{}}>
+            <div>
+              <Text>Bundle is missing required ValueSets.</Text>
+              {error.isThrownFromMrs && (
+                <Text>
+                  A VS API Key must be specified by the Measure Repository Service server to retrieve the missing
```
sarahmcdougall, line 31: I think this addition is very helpful! :)

sarahmcdougall (changes requested):

components/measure-upload/MeasureRepositoryUpload.tsx
elsaperelli: Wait actually I am having a hard time replicating this error

sarahmcdougall (approved):

PR #65, opened 2023-05-01
Add Full URL to Export Bundles

natjoe4:
# Summary
Adds a `fullUrl` to each exported bundle entry. Preserves `fullUrls` from imported bundles
Open Questions: 
- Should we preserve the `fullUrl` for imported testCaseMeasureReports?
- Should we update the `fullUrl` if the id of an existing resource is updated?

## New Behavior
When a patient test case is exported, each entry in the bundle will now have a `fullUrl` field. This `fullUrl` will be preserved if the patient test case was initially an imported case that had `fullUrls` defined, or will be generated in the `urn:uuid` style

## Code Changes
 - Updated `PatientTestCase` to store a fullUrl for the Patient and an array of BundleEntries instead of resources to preserve `fullUrl` field
 - Updated `createPatientBundle` to take in a `fullUrl` for the patient and add `fullUrl`s to the bundle entries
 - Updated all components that access resources to point to the nested `entry.resource`
 - Updated patient cloning to update `fullUrl`s for cloned patient as well
 - Updated testing 

# Testing Guidance
 - `npm run check`
 - `npm run dev`
 - Upload a measure and create a few new patient test cases, each with a few resources
 - Export a single test case and ensure that the exported bundle has `fullUrl` fields for all the bundle entries
 - Clone a patient
 - Export all test cases and ensure they all have `fullUrl` fields. Also ensure that the cloned test case has appropriate `fullUrl`s that match the ids of the cloned resources.
 - Restart and import one of your patient test cases
 - Now export the test case again and ensure the `fullUrl`s match from the import

### DIFF
--- a/__tests__/components/calculation/PopulationCalculation.test.tsx
+++ b/__tests__/components/calculation/PopulationCalculation.test.tsx
@@ -80,6 +80,7 @@ describe('PopulationCalculation', () => {
           resourceType: 'Patient',
           name: [{ given: ['Test123'], family: 'Patient456' }]
         },
+        fullUrl: 'urn:uuid:testPatient',
         resources: []
       }
     });
@@ -105,6 +106,7 @@ describe('PopulationCalculation', () => {
           resourceType: 'Patient',
           name: [{ given: ['Test123'], family: 'Patient456' }]
         },
+        fullUrl: 'urn:uuid:testPatient',
         resources: [
           {
             resource: {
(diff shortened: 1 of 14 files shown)


hossenlopp:
With those two open questions, yes and yes. Updating the fullUrls with the new ids is especially important when updating them or cloning a patient or resource.

sarahmcdougall (changes requested):
Great work! I agree with Hoss's response in regard to those open questions. Aside from that, I have just a few suggested changes

hossenlopp (approved):
Works well! Am able to import into MADiE without it getting upset about lack of fullURLs.

sarahmcdougall (approved):
Great changes! And good discussion on and off Github regarding the details of updating the `fullUrl` :)

PR #67, opened 2023-05-04
Observation results

lmd59:
# Summary

## Code Changes

# Testing Guidance


util/MeasurePopulations.ts
```diff
@@ -33,16 +29,41 @@ export function getMeasurePopulationsForSelection(measure: fhir4.Measure): Multi
...
   });
   return measurePopulations;
 }
 
+/**
+ * Creates a label string that can be attached to the default (population criteria, etc.) and adjusts to show measure
+ * observation related population reference so that the population string is unique across different measure observations
+ * @param {fhir4.MeasureGroupPopulation} population - FHIR population from a measure group
+ * @param {fhir4.MeasureGroup} group - FHIR measure group for observation population lookup
+ * @returns {string} adjustment string
```
elsaperelli, line 49: I feel like these are a bit unnecessary but up to you!

util/MeasurePopulations.ts
```diff
@@ -33,16 +29,41 @@ export function getMeasurePopulationsForSelection(measure: fhir4.Measure): Multi
...
+  population: fhir4.MeasureGroupPopulation,
+  group: fhir4.MeasureGroup | undefined
+): string {
+  let label = ''; // default adjustment is none
+  const criteriaReference = population?.extension?.find(e => {
+    return e?.url === 'http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/cqfm-criteriaReference';
+  })?.valueString;
+  // handle measure observation
+  if (group && population.code?.coding?.[0].code === 'measure-observation' && criteriaReference) {
+    const obsPop = group.population?.find(p => p && p.id === criteriaReference);
```
elsaperelli, line 61:
```suggestion
    const obsPop = group.population?.find(p => p.id === criteriaReference);
```
I tested it and it seems to work?

elsaperelli (changes requested):
Great collab! I have some small inline comments and one question so far.

I uploaded the `ratio-Encounter-reuseObservationFunction-bundle.json` and just created a new patient. I added some desired populations, but the population comparison table is blank:

image

This is the behavior with the same measure on the main branch:

image


We probably want this case to still show the values even if the patient didn't calculate into the ipp, right? Happy to discuss more!

components/calculation/PopulationComparisonTable.tsx
```diff
@@ -47,45 +47,193 @@ export default function PopulationComparisonTable({ patientId }: PopulationCompa
...
+
+  /**
+   * Creates an array of ResultValues objects for each of the Episode results. Specifies resource (which episode),
+   * desired population values, and actual population values. Desired and actual are key/value pairs,
+   * where the population label (criteria) is the key and the value is 0/1 or the observation value if the population
+   * is a measure observation.
+   */
+  function constructEpisodeValues(group: DetailedPopulationGroupResult | undefined) {
+    const results = group?.episodeResults?.map(er => {
+      const episode = currentPatients[patientId].resources.find(r => r.id === er.episodeId)?.resource;
```
elsaperelli, line 150:
```suggestion
      const episode = currentPatients[patientId].resources.find(r => r.resource?.id === er.episodeId)?.resource;
```

When I upload the measure bundle from the fqm-execution integration test file and import a patient from that file, the episode result labels are undefined: undefined: 

image


I did some quick digging and I think it's because id isn't accessed on the resource but actually r.resource.id.

elsaperelli: image

That seemed to fix the last three

elsaperelli: I think there should only be three though...

elsaperelli (changes requested):
I found some other interesting behavior that I am a bit confused by. Apologies my comments are all over the place- would love to discuss!

components/calculation/PopulationComparisonTable.tsx
```diff
@@ -30,11 +31,10 @@ export interface PopulationComparisonTableProps {
 export interface DesiredPopulations {
   [key: string]: number;
 }
-export interface BothPopulations {
-  [key: string]: {
-    desired: number;
-    actual: number | undefined;
-  };
+export interface ResultValues {
```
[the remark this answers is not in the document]
lmd59: @natjoe4 made the great suggestion to use Record!

README.md
```diff
@@ -95,6 +98,16 @@ FQM-Testify allows the user to create non-patient FHIR resources after at least
...
 
+#### Reading the Population Comparison Table
+The columns in this table represent the possible populations as specified on the uploaded Measure and the rows represent the desired and actual populations as stored/calculated by FQM-Testify. The first rows are for the patient, and any subsequent rows show episode results.
+Both patient and episode results compare actual and desired results. If any of the cells are highlighted red, that means the desired population/observation does not match the actual population/observation. If they are highlighted green, then they match. If they are not highlighted, then the cell represents a value that is not relevant to the current patient/measure.
+
+##### Proportion
+Proportion measures show only patient population results. For the &apos;Actual&apos; row, the value will be 1 if the patient calculates into the respective population, and 0 if it does not. The &apos;Desired&apos; row shows a 1 for populations that the patient are expected to be a part of and 0 for populations that the patient is not expected to be a part of.
+
+##### Ratio
```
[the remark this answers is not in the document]
lmd59: Hmm- I think this is a bit of a historical artifact of expectations. I would defer to @hossenlopp on this one.

components/calculation/PopulationComparisonTable.tsx
```diff
@@ -47,45 +47,193 @@ export default function PopulationComparisonTable({ patientId }: PopulationCompa
     return measureBundle.content?.entry?.find(e => e.resource?.resourceType === 'Measure')?.resource as fhir4.Measure;
   }, [measureBundle]);
 
+  /**
+   * Creates a key that can be used to represent the population associated with a particular population result.
+   * This key can be used to show the user what population the results are relevant for. It generally uses the
+   * criteria expression, but makes adjustments for measure observations where the expression may not be unique.
+   */
+  function keyForResult(result: PopulationResult, group: DetailedPopulationGroupResult | undefined) {
```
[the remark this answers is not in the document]
lmd59: Updated the import to import directly from `fqm-execution`. Is that what you meant?

components/calculation/PopulationComparisonTable.tsx
lmd59: Is this resolved at this point? Not sure if you were still seeing issues after the [Fix resource name issue (finish rebase fix)](https://github.com/projecttacoma/fqm-testify/pull/67/commits/f481be6e4ce466893b2f7fc0394cca01f3d91b28) update?

components/calculation/PopulationComparisonTable.tsx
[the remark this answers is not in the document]
lmd59: Actually- my import update caused an issue in ci... but seemed to work in local. Maybe we hold off for now...? heh.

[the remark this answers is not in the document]
elsaperelli: Yes probably better to hold off for now.

elsaperelli (changes requested):
image

It looks like when you add desired popoulations to a patient, those are not reflected in the population comparison table. For example, in the above screen, the patient's desired populations are ipp, denom, and numer, but the population table has 0 for all of them.

lmd59:
> image
> 
> It looks like when you add desired populations to a patient, those are not reflected in the population comparison table. In the readme, I think this makes sense because now the patient results are for the total number of episode results in the given population? Will we add functionality for desired populations for episode based measures in the future or should we honestly disable the desired populations picker for episode based measures?

Yes- agreed, it does not update population information for episode-based measures because the current desired population selection does not align with this. There are TODOs in the code to update the default values once there's a way to set the desired population. Currently, we do have a ticket for being able to set these desired values (TACOSTRAT-1283). Not 100% sure when we're getting to it - maybe next sprint? Could ask Sarah/Matt about this.

components/calculation/PopulationComparisonTable.tsx
```diff
@@ -47,45 +47,195 @@ export default function PopulationComparisonTable({ patientId }: PopulationCompa
...
+    let key = result.criteriaExpression as string;
```
hossenlopp, line 56: I would still suggest not using the criteria expression and suggest using the `populationType` mapped to a short hand. ex. `initial-population` -> `IPP`.  It is decently common for measures to use the same define statement for both IPP and DENOM. So in this situation you would see "Initial Population" as the header for two columns.

hossenlopp (changes requested):
I few small comments on comments and a suggestion to make the table consistent for the population types.

components/calculation/PopulationComparisonTable.tsx
```diff
@@ -47,45 +47,195 @@ export default function PopulationComparisonTable({ patientId }: PopulationCompa
...
+   * Creates an array of ResultValues objects for each of the Episode results. Specifies resource (which episode),
+   * desired population values, and actual population values. Desired and actual are key/value pairs,
+   * where the population label (criteria) is the key and the value is 0/1 or the observation value if the population
+   * is a measure observation.
+   */
+  function constructEpisodeValues(group: DetailedPopulationGroupResult | undefined) {
+    const results = group?.episodeResults?.map(er => {
+      const episode = currentPatients[patientId].resources.find(r => r.resource?.id === er.episodeId)?.resource;
+      const keys = er.populationResults.map(pr => keyForResult(pr, group));
+      // dummy placeholder until episodes have desired results: all values currently 0
```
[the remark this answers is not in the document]
lmd59: Hah- just added the "TODO" on this in the commit I just pushed. It's a relatively small commit - actually changed the placeholder to -1 and completely disabled the desired rows for episode measures.

components/calculation/PopulationComparisonTable.tsx
hossenlopp: I think OBSERV, OBSERV-NUMER, OBSERV-DENOM. would work well

hossenlopp (approved):
Sweet! Can see individual episode results. This is huge!

util/MeasurePopulations.ts
```diff
@@ -6,43 +6,72 @@ export interface MultiSelectData {
...
+  denominator: 'DENOM',
+  'denominator-exclusion': 'DENEX',
+  'denominator-exception': 'DENEXCEP',
+  numerator: 'NUMER',
+  'numerator-exclusion': 'NUMEX',
+  'measure-population': 'MSRPOPL',
+  'measure-population-exclusion': 'MSRPOPLEX',
+  'measure-observation': 'OBSERV'
+};
+
```
elsaperelli, line 20: Rather than adding this const, I think can use the PopulationType enum exported from fqm-execution. What are your thoughts on this?

elsaperelli (changes requested):

util/MeasurePopulations.ts
mgramigna: This might be a little tough to do in code and might even be impossible, I'd maybe just leave this as is

lmd59: I think this kind of reverse mapping to get the Enum name is possible for numeric enums (https://www.typescriptlang.org/docs/handbook/enums.html#reverse-mappings), but I'm not seeing an equivalent for string enums. @elsaperelli - okay with giving up on making this change?

hossenlopp (approved):

PR #68, opened 2023-05-11
Searchable Expressions

natjoe4:
# Summary
Add an autocomplete input that scrolls the highlighted cql to the selected expression def

## New Behavior
New autocomplete input under the population results table that scrolls to the input cql expression definition in the highlighted cql

## Code Changes
 - added HTML ids to each define statement (identified by regex) during the parse HTML step
 - added an autocomplete input component
 - updated testing to reflect new components

# Testing Guidance
 - `npm run check`
 - Fire up testify and input a MeasureBundle
 - Create a patient and select it. Ensure that the autocomplete input is populated with all valid expression defs
 - Click on an option and watch it scroll!

### DIFF
--- a/components/calculation/PopulationComparisonTable.tsx
+++ b/components/calculation/PopulationComparisonTable.tsx
@@ -4,7 +4,7 @@ import { patientTestCaseState } from '../../state/atoms/patientTestCase';
 import { measureBundleState } from '../../state/atoms/measureBundle';
 import { useMemo, useState } from 'react';
 import { getMeasurePopulationsForSelection, MultiSelectData, getPopShorthand } from '../../util/MeasurePopulations';
-import { InfoCircle } from 'tabler-icons-react';
+import { InfoCircle, Search } from 'tabler-icons-react';
 import { detailedResultLookupState } from '../../state/atoms/detailedResultLookup';
 import { DetailedPopulationGroupResult, PopulationResult } from 'fqm-execution/build/types/Calculator';
 import React from 'react';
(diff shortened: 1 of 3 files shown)


components/calculation/PopulationComparisonTable.tsx
```diff
@@ -37,12 +38,14 @@ export interface BothPopulations {
...
 }
 
-export default function PopulationComparisonTable({ patientId }: PopulationComparisonTableProps) {
+export default function PopulationComparisonTable({ patientId, defIds }: PopulationComparisonTableProps) {
   const { classes } = useStyles();
   const measureBundle = useRecoilValue(measureBundleState);
   const detailedResultLookup = useRecoilValue(detailedResultLookupState);
   const currentPatients = useRecoilValue(patientTestCaseState);
   const [opened, setOpened] = useState(false);
+  const [searchValue, setSearchValue] = useState<string>('');
```
elsaperelli, line 47:
```suggestion
  const [searchValue, setSearchValue] = useState('');
```

elsaperelli, line 47:
```suggestion
  const [searchValue, setSearchValue] = useState('');
```

components/calculation/PopulationComparisonTable.tsx
```diff
@@ -145,6 +148,15 @@ export default function PopulationComparisonTable({ patientId }: PopulationCompa
             })}
           </tbody>
         </Table>
+        <Autocomplete
+          data={Object.keys(defIds)}
```
elsaperelli, line 152: Is it possible to make the defIds that pop up scrollable? When I first click on the input the following are displayed but in order to see the others, I have to start typing:

image

elsaperelli (changes requested):
Great work! This will be super helpful. Just two small inline comments.

elsaperelli (approved):
Incredible

PR #73, opened 2023-06-01
Collapsible table

lmd59:
# Summary
The population comparison table in the measure highlighting panel is collapsible (note: this is an update from the ticket to just collapse the table and not collapse the html markup, based on a conversation in dev hour)

## New Behavior
- The user may click on the population table title area to collapse the table. This uses an accordion style component to help with user-familiar iconography, but the accordion only applies to the one element for now.

## Code Changes
- `PopulationComparisonTable.tsx` - Move clause search to `MeasureHighlightingPanel`, move table title and popover tooltip to their own function so that they can be used as the Accordion control bar, stop popover from clicking through to control bar with updated popoverClick function
- `MeasureHighlightingPanel.tsx` - Add clause search (moved from `PopulationComparisonTable`, use mantine `Accordion` component for collapsible table, use table title and popover as part of accordion control
- `PopulationComparisonTable.test.tsx` - simple change to get rid of an unnecessary pass to the PopulationComparisonTable function component (not needed now that clause search is removed from the table component)

# Testing Guidance
- `npm run check`
- `npm run dev`
- Try uploading different measure bundles and patients and see how it looks. This will also be influenced by scrollable areas once #72 is merged
- Potential suggested test cases: 
  - `CervicalCancerScreeningFHIR-bundle` measure bundle and patients from the ecqm-content-r4 measures. This will create a smaller table.
  - `ratio-Encounter-reuseObservationFunction` measure bundle and patients (especially `patient-3enc-1in-all-2in-denom`) from fqm-execution `test/integration` folder. This will create a larger table.

### DIFF
--- a/components/calculation/MeasureHighlightingPanel.tsx
+++ b/components/calculation/MeasureHighlightingPanel.tsx
@@ -1,11 +1,10 @@
-import { Accordion, Autocomplete, Button, Collapse, ScrollArea, Space, Text, createStyles } from '@mantine/core';
+import { Accordion, Autocomplete, ScrollArea, Space, Text, createStyles } from '@mantine/core';
 import { useRecoilValue } from 'recoil';
 import parse from 'html-react-parser';
 import PopulationComparisonTable, { PopulationComparisonTableControl } from './PopulationComparisonTable';
 import { detailedResultLookupState } from '../../state/atoms/detailedResultLookup';
 import { useMemo, useState } from 'react';
 import { Text as DomText } from 'domhandler';
-import { useDisclosure } from '@mantine/hooks';
 import { Search } from 'tabler-icons-react';
 
 /**
--- a/components/calculation/MeasureHighlightingPanel.tsx
+++ b/components/calculation/MeasureHighlightingPanel.tsx
@@ -1,11 +1,12 @@
 import { Accordion, Autocomplete, ScrollArea, Space, Text, createStyles } from '@mantine/core';
 import { useRecoilValue } from 'recoil';
 import parse from 'html-react-parser';
-import PopulationComparisonTable, { PopulationComparisonTableControl } from './PopulationComparisonTable';
+import PopulationComparisonTable from './PopulationComparisonTable';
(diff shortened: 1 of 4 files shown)


components/calculation/PopulationComparisonTable.tsx
```diff
@@ -313,28 +273,54 @@ export default function PopulationComparisonTable({ patientId, defIds }: Populat
...
-          }}
-        />
       </>
     );
   } else {
     return <Text>Measure highlighting not available</Text>;
   }
 }
+
+export function PopulationComparisonTableControl() {
```
mgramigna, line 283: Convention is usually to not put 2 components in the same file especially if one doesn't render the other. Can you move `PopulationComparisonTableControl` to its own file?

components/calculation/MeasureHighlightingPanel.tsx
```diff
@@ -48,9 +49,41 @@ export default function MeasureHighlightingPanel({ patientId }: MeasureHighlight
     return { parsedHTML, defIds };
   }, [detailedResultLookup, patientId]);
 
+  // handle search
+  const [searchValue, setSearchValue] = useState('');
+
   return (
     <>
-      <PopulationComparisonTable patientId={patientId} defIds={defIds} />
+      <Accordion chevronPosition="left" defaultValue="table">
```
mgramigna, line 57: I feel like it make more sense to use `Collapse` instead of Accordion since usually Accordions are for things with multiple items that you want to toggle between. However, that's more work, so this is fine

elsaperelli (approved):
I agree with Matt's comments but otherwise lgtm

components/calculation/MeasureHighlightingPanel.tsx
lmd59: Yeah- I ended up going this direction because basically all the good examples I could find for collapsible components were styled in this way, and I wanted to rely on user familiarity with existing component styles. I think this makes it look usable in the right way and also lets us easily work with collapsing individual pieces of the table in the future if we want to do that.

mgramigna (approved):

PR #70, opened 2023-05-23
Update to Mantine 6 and Next 13

mgramigna:
# Summary

Upgrades the two key dependencies, making necessary small changes.

## New Behavior

None, except for measurement period entry is two [DateInput](https://mantine.dev/dates/date-input/) components (new in Mantine 6)

## Code Changes

* Update dependencies, and purge unused dependencies (fhir-typescript, emotion cache)
* Deleted a test that no longer works due to underlying changes in the mantine select component (lame)
* Separate MP start and end into separate atoms and change the date inputs
* Update props to adhere to mantine 6 and removed some no longer supported props

# Testing Guidance

Test all features. Make sure everything works the same as on https://projecttacoma.github.io/fqm-testify/

PR #75, opened 2023-06-08
Refactor existing UI/UX

mgramigna:
# Summary

This PR makes some sweeping UI changes across the board that can loosely be categorized as follows:

1. Generally avoid use of inline `style` prop where not necessary, either by moving things out to `createStyles` or using a proper Mantine component
2. Remove some unnecessary Grid systems, and other unnecessary styles that were either doing nothing or were causing weird behavior like double scrollbars
3. Clean up the homepage and make it more responsive to all screen sizes
4. Bug fixes or misc. style changes (e.g. text emphasis, padding, etc.)

## New Behavior

* Bug fix for the patient cards now properly highlighting when selected, which they were no longer doing on `main` after #70 

No other new behavior beyond that

## Code Changes

* Yoink a lot of inline `style` to either `createStyles`, `styles`, or `sx`, which is more akin to what you should be doing in any full-fledged component library like mantine
* Replace many `div`s or `Grid`s with flexbox where appropriate
  * Also replace some misc tags with Mantine components (e.g. `<i>some text</>` to `<Text italic>some text</Text>`
* Fix error introduced by #73 which caused invalid dom nesting (button within a button)
  * To do this, I followed the official suggestion for this behavior made by the mantine docs: https://mantine.dev/core/accordion/#compose-controls
* Rework homepage spacing and layout by using specific grids and max widths on large screens
* Update tests to adhere to some wording changes in the UI 

# Testing Guidance

Test all features, test various screen sizes


elsaperelli (changes requested):
Just one comment! Great changes!

components/calculation/PopulationComparisonTableControl.tsx
```diff
@@ -12,40 +12,31 @@ export default function PopulationComparisonTableControl() {
...
-        </Popover>
-      </div>
-    </Group>
+    <Popover opened={opened} onClose={() => setOpened(false)} width={500}>
+      <Popover.Target>
+        <ActionIcon aria-label={'More Information'} onClick={popoverClick}>
+          <InfoCircle size={20} />
+        </ActionIcon>
+      </Popover.Target>
+      <Popover.Dropdown>
```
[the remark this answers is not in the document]
mgramigna: Yeah, I noticed this too. It's also like this on `main`, so I just didn't change that behavior 
image

components/calculation/PopulationComparisonTableControl.tsx
```diff
@@ -12,40 +12,31 @@ export default function PopulationComparisonTableControl() {
...
+        <ScrollArea h={180}>
+          The Population Comparison Table shows patient and episode population results for the patient selected. For
+          patient-based measures, patient results show 0 or 1 to indicate belonging to a population. Actual and desired
+          populations are compared to highlight cells green if they match and red if they don&apos;t match.
```
sarahmcdougall, line 25:
```suggestion
          populations are compared, and the cells are highlighted green if the populations match and red if they don&apos;t match.
```

components/calculation/PopulationComparisonTableControl.tsx
```diff
@@ -12,40 +12,31 @@ export default function PopulationComparisonTableControl() {
...
+          <br />
+          <br />
+          For episode-based measures, the table shows patient level totals that indicate how many episodes are in each
```
sarahmcdougall, line 28:
```suggestion
          For episode-based measures, the table shows patient-level totals that indicate how many episodes are in each
```

components/TestCaseEditor.tsx
```diff
@@ -69,42 +71,52 @@ export default function TestCaseEditor() {
...
+                  {isCalculationLoading ? (
+                    <Center>
+                      <Loader size={40} />
+                      <Text italic color="dimmed">
+                        Calculating...
+                      </Text>
+                    </Center>
+                  ) : (
+                    <Center>
+                      <CircleCheck color="green" size={40} />
```
sarahmcdougall, line 95: Really minor suggestion here but this circle check looks pretty large compared to the size of the text - I would probably bump the size down to like 28 or 32-ish

sarahmcdougall (changes requested):
Great changes - this really enhances the UI and the readability of our code. Just have a few suggestions

sarahmcdougall (approved):
Great changes! 👍

PR #77, opened 2023-07-12
Pretty highlighting

elsaperelli:
# Summary
This PR adds rendering of pretty statement results in the measure highlighting panel.

## New Behavior
Each statement result in the `MeasureHighlightingPanel` now has a button below it that says "Show Results". When the user clicks on this button, a collapsible component is opened that contains the pre-formatted pretty statement results. The `Autocomplete` component behaves the same way, it was just updated to utilize the `data-statement-name` in the html rather than a defined Regular Expression.

## Code Changes
- `MeasureHighlightingPanel.tsx` - Parses through the html and inserts a `PrettyOutput` component below each statement result in the measure logic highlighting. It also updates the `Autocomplete` component to use `data-statement-name` rather than the Regular Expression and the `defIds` record. The `Autocomplete` data attribute is populated with all of the statement named in the statement results that have relevance !== null (so either TRUE or FALSE) because these are the statements that are rendered in the highlighting.
- `PrettyOutput.tsx` - New component that consists of a Button and a collapsible component that contains the statement result pretty output.

# Testing Guidance
- `npm run check`
- `npm run dev`
- Upload a measure and check out a patient's measure highlighting panel. Test the buttons, make sure the pretty results are correct, let me know if you have any styling suggestions!

### DIFF
--- a/components/calculation/MeasureHighlightingPanel.tsx
+++ b/components/calculation/MeasureHighlightingPanel.tsx
@@ -31,8 +31,8 @@ export default function MeasureHighlightingPanel({ patientId }: MeasureHighlight
     const parsedHTML = parse(detailedResultLookup[patientId]?.detailedResults?.[0].html || '', {
       replace: elem => {
         if ((elem as DomElement).attribs?.['data-statement-name']) {
-          const statementName = (elem as DomElement).attribs?.['data-statement-name'];
-          const libraryName = (elem as DomElement).attribs?.['data-library-name'];
+          const statementName = (elem as DomElement).attribs['data-statement-name'];
+          const libraryName = (elem as DomElement).attribs['data-library-name'];
           const statementResult = detailedResultLookup[patientId]?.detailedResults?.[0].statementResults.find(
             statement => statement.statementName === statementName && statement.libraryName === libraryName
           );
--- a/components/calculation/MeasureHighlightingPanel.tsx
+++ b/components/calculation/MeasureHighlightingPanel.tsx
@@ -32,8 +32,9 @@ export default function MeasureHighlightingPanel({ patientId }: MeasureHighlight
       replace: elem => {
         if ((elem as DomElement).attribs?.['data-statement-name']) {
           const statementName = (elem as DomElement).attribs?.['data-statement-name'];
+          const libraryName = (elem as DomElement).attribs?.['data-library-name'];
           const statementResult = detailedResultLookup[patientId]?.detailedResults?.[0].statementResults.find(
-            statement => statement.statementName === statementName
+            statement => statement.statementName === statementName && statement.libraryName === libraryName
           );
           return (
             <>
--- a/components/calculation/MeasureHighlightingPanel.tsx
+++ b/components/calculation/MeasureHighlightingPanel.tsx
@@ -73,7 +73,7 @@ export default function MeasureHighlightingPanel({ patientId }: MeasureHighlight
         onItemSubmit={item => {
           document
             .querySelector(
-              `pre[data-statement-name="${item.value.split('"')[1]}"[data-library-name="${item.value.split('"')[0]}"]`
+              `pre[data-statement-name="${item.value.split('"')[1]}"][data-library-name="${item.value.split('.')[0]}"]`
             )
             ?.scrollIntoView({ behavior: 'smooth' });
         }}
--- a/components/calculation/MeasureHighlightingPanel.tsx
+++ b/components/calculation/MeasureHighlightingPanel.tsx
@@ -72,7 +72,9 @@ export default function MeasureHighlightingPanel({ patientId }: MeasureHighlight
         label="Search CQL Expression Definition"
         onItemSubmit={item => {
           document
-            .querySelector(`[data-statement-name="${item.value.split('"')[1]}"`)
+            .querySelector(
+              `pre[data-statement-name="${item.value.split('"')[1]}"[data-library-name="${item.value.split('"')[0]}"]`
+            )
             ?.scrollIntoView({ behavior: 'smooth' });
         }}
         rightSection={
--- a/components/calculation/MeasureHighlightingPanel.tsx
+++ b/components/calculation/MeasureHighlightingPanel.tsx
@@ -1,16 +1,11 @@
-import { Autocomplete, ScrollArea, Space, Text, createStyles } from '@mantine/core';
+import { ActionIcon, Autocomplete, ScrollArea, Space, Text, createStyles } from '@mantine/core';
 import { useRecoilValue } from 'recoil';
-import parse from 'html-react-parser';
+import parse, { domToReact } from 'html-react-parser';
 import { detailedResultLookupState } from '../../state/atoms/detailedResultLookup';
 import { useMemo, useState } from 'react';
-import { Text as DomText } from 'domhandler';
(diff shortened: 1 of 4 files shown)


sarahmcdougall (changes requested):
Great job implementing this! In addition to in-line suggestions, I also suggest running `npm upgrade fqm-execution` to update the package-lock commit hash so that it's up-to-date

components/calculation/PrettyOutput.tsx
```diff
@@ -0,0 +1,28 @@
...
+  statement: StatementResult | undefined;
+}
+
+export default function PrettyOutput({ statement }: PrettyOutputProps) {
+  const [opened, { toggle }] = useDisclosure(false);
+  const buttonText = opened ? 'Hide Results' : 'Show Results';
+
+  return (
+    <>
+      <div style={{ width: 120 }}>
```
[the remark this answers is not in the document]
elsaperelli: Good call!

components/calculation/PrettyOutput.tsx
```diff
@@ -0,0 +1,28 @@
...
+        <Button fullWidth compact variant="outline" color="gray" onClick={toggle}>
+          {buttonText}
+        </Button>
+      </div>
+      <Collapse in={opened}>
+        <Space />
```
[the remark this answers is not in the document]
elsaperelli: Yeah I also don't think this was doing anything so I just removed it, but let me know if I should add any more padding/margins.

components/calculation/MeasureHighlightingPanel.tsx
```diff
@@ -68,8 +70,19 @@ export default function MeasureHighlightingPanel({ patientId }: MeasureHighlight
         limit={100}
         label="Search CQL Expression Definition"
         onItemSubmit={item => {
-          document.getElementById(defIds[item.value])?.scrollIntoView({ behavior: 'smooth' });
+          document
+            .querySelector(`[data-statement-name="${item.value.split('"')[1]}"`)
```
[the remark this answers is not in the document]
elsaperelli: I don't think we need to since it'll look for that string, but I can add it if you think it's necessary?

components/calculation/MeasureHighlightingPanel.tsx
```diff
@@ -32,28 +27,35 @@ export default function MeasureHighlightingPanel({ patientId }: MeasureHighlight
...
-          const data = (elem as DomText).data;
-          if (data.match(expressionDefRegex)) {
-            const splitData = data.split('"');
-            defIds[splitData[1]] = splitData[1].toLowerCase().replace(' ', '-');
-            return <span id={defIds[splitData[1]]}>{data}</span>;
-          }
+        if ((elem as DomElement).attribs?.['data-statement-name']) {
+          const statementName = (elem as DomElement).attribs?.['data-statement-name'];
+          const statementResult = detailedResultLookup[patientId]?.detailedResults?.[0].statementResults.find(
+            statement => statement.statementName === statementName
```
mgramigna, line 36: need to also check the value of `attribs['data-library-name']` and use that in the `.find` for a statement result since that will guarantee the uniqueness. Just statement name won't

components/calculation/MeasureHighlightingPanel.tsx
```diff
@@ -32,28 +27,36 @@ export default function MeasureHighlightingPanel({ patientId }: MeasureHighlight
...
-        if (elem.type === 'text') {
-          const data = (elem as DomText).data;
-          if (data.match(expressionDefRegex)) {
-            const splitData = data.split('"');
-            defIds[splitData[1]] = splitData[1].toLowerCase().replace(' ', '-');
-            return <span id={defIds[splitData[1]]}>{data}</span>;
-          }
+        if ((elem as DomElement).attribs?.['data-statement-name']) {
+          const statementName = (elem as DomElement).attribs?.['data-statement-name'];
+          const libraryName = (elem as DomElement).attribs?.['data-library-name'];
```
sarahmcdougall, line 35: Since we do a check in the `if` statement that `attribs` is defined, could we change these to `(elem as DomElement).attribs['data-statement-name']` and `(elem as DomElement).attribs['data-library-name']`?

elsaperelli: Nice catch!

components/calculation/MeasureHighlightingPanel.tsx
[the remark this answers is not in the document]
hossenlopp: The query selector we really want to create is: `pre[data-statement-name="statement Name"][data-library-name="library name"]`

Adding the `pre` on the front makes it more clear what we are looking for and potentially faster to. We definitely need the closing square brackets to make it a valid CSS query selector since some browsers may choose to completely throw out selectors that have any parsing issues.

components/calculation/PrettyOutput.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { useDisclosure } from '@mantine/hooks';
+import { StatementResult } from 'fqm-execution';
+
+export interface PrettyOutputProps {
+  statement: StatementResult | undefined;
+}
+
+export default function PrettyOutput({ statement }: PrettyOutputProps) {
+  const [opened, { toggle }] = useDisclosure(false);
+  const buttonText = opened ? 'Hide Results' : 'Show Results';
```
hossenlopp, line 11: Just a question. Why put the button text here instead of in the TSX stuff below?

hossenlopp (changes requested):
This works really well. Even stays open and not scrolled on changes to the patient and updates appropriately. 

Just have a single suggestion and question on one thing. Additionally `fqm-execution` needs to be updated to make this PR work properly. Run `npm update fqm-execution` and check in the `package-lock.json` changes. This will allow us to have it working in the GH pages deploy immediately.

components/calculation/PrettyOutput.tsx
elsaperelli: Honestly, I am not sure 😆 But I am going to move it in the TSX stuff below because I agree that makes more sense.

components/calculation/MeasureHighlightingPanel.tsx
```diff
@@ -68,8 +71,21 @@ export default function MeasureHighlightingPanel({ patientId }: MeasureHighlight
         limit={100}
         label="Search CQL Expression Definition"
         onItemSubmit={item => {
-          document.getElementById(defIds[item.value])?.scrollIntoView({ behavior: 'smooth' });
+          document
+            .querySelector(
+              `pre[data-statement-name="${item.value.split('"')[1]}"[data-library-name="${item.value.split('"')[0]}"]`
```
hossenlopp, line 76: This still needs the `]` for the first part of the selector on the statement name. Currently this makes chrome throw an error. You can see in this case right after `"Denominator"` it needs the `]`

```
Uncaught DOMException: Failed to execute 'querySelector' on 'Document': 'pre[data-statement-name="Denominator"[data-library-name="HIVScreeningFHIR."]' is not a valid selector.
```

hossenlopp (changes requested):
Selector needs a fix on the jump to statement functionality.

components/calculation/MeasureHighlightingPanel.tsx
hossenlopp: Also just noticed that you will not want the `.` in the library name.

PR #79, opened 2023-09-15
Add trust meta profile option

elsaperelli:
# Summary
- This PR adds a UI component that allows the user to specify whether or not they want to run calculation with `trustMetaProfile`. `trustMetaProfile` is an option in `fqm-execution` and is defaulted to false. 

## New Behavior
- The user can now opt to set `trustMetaProfile` to true through the UI.

## Code Changes
- `README.md` - added short sentence about `trustMetaProfile` (is this enough?)
- `PopulationCalculation.tsx` - set `calculateSDEs` to `true` and `trustMetaProfile` to the value of the `trustMetaProfile` atom in the CalculationOptions.
- `PatientCreationPanel.tsx`/`ResourceDisplay.tsx`/`ResourcePanel.tsx`/`generate-test-cases.tsx` - pass `trustMetaProfile` atom value into calls to `calculateDetailedResult()`
- `MeasureFileUploadHeader.tsx` - moved popover to be next to the "Step" for consistency
- `package.json`/`package-lock.json` - upgraded fqm-execution version.
- `index.tsx` - Added switch component for `trustMetaProfile`
- `trustMetaProfile.ts` - created an atom for `trustMetaProfile`
- `MeasureCalculation.ts` - added `trustMetaProfile` as a parameter to `calculateDetailedResult()`

# Testing Guidance
- `npm run check`
- `npm run dev`
- Navigate to `localhost:3000`
- Upload `CMS1028` measure bundle (one example) and make sure the `useTrustMetaProfile` switch is off.
- Upload the associated test cases and click "Calculate Population Results". An error from cql-execution should pop up.
- Repeat these steps but set `useTrustMetaProfile` to yes. There should no longer be an error from cql-execution.
- Do we like the switch component for this? Would it make more sense to have a check box? Any other suggestions?


components/calculation/PopulationCalculation.tsx
```diff
@@ -24,6 +25,7 @@ export default function PopulationCalculation() {
   const [enableTableButton, setEnableTableButton] = useState(false);
   const [enableClauseCoverageButton, setEnableClauseCoverageButton] = useState(false);
   const [clauseCoverageHTML, setClauseCoverageHTML] = useState<string | null>(null);
+  const useTrustMetaProfile = useRecoilValue(trustMetaProfile);
```
hossenlopp, line 28: Should this just be `trustMetaProfile` for consistency with the atom name? Or perhaps the atom name could be `useTrustMetaProfile` instead.

components/patient-creation/PatientCreationPanel.tsx
```diff
@@ -82,7 +84,8 @@ function PatientCreationPanel() {
               currentPatients[id],
               measureBundle.content,
               measurementPeriod.start?.toISOString(),
-              measurementPeriod.end?.toISOString()
+              measurementPeriod.end?.toISOString(),
+              useTrustMetaProfile
```
hossenlopp, line 88: Same comment as elsewhere, should this match the atom name?

hossenlopp (changes requested):
Works great. The switch could be moved into that Step 2 section as it is an option for calculation like the measurement period is.

I think the naming and usage of the atom for the setting could be match what we have done for other atoms. I commented in a few places my thought on this.

state/atoms/trustMetaProfile.ts
```diff
@@ -0,0 +1,9 @@
+import { atom } from 'recoil';
+
+/**
+ * Atom indicating if trustMetaProfile should be set to true
+ */
+export const trustMetaProfile = atom<boolean>({
+  key: 'trustMetaProfile',
```
hossenlopp, line 7:
```suggestion
export const trustMetaProfileState = atom<boolean>({
  key: 'trustMetaProfileState',
```
Should both of these be `trustMetaProfileState` to match what we named other atoms?

components/calculation/PopulationCalculation.tsx
elsaperelli: I get an error if they are the same, but since I changed the `trustMetaProfile` atom to `trustMetaProfileState`, I will change this line to `const trustMetaProfile = useRecoilValue(trustMetaProfileState)`. Does that make sense? Or should it be something else?

components/resource-creation/ResourceDisplay.tsx
```diff
@@ -96,7 +96,7 @@ function ResourceDisplay() {
           measureBundle.content,
           measurementPeriod.start?.toISOString(),
           measurementPeriod.end?.toISOString(),
-          useTrustMetaProfile
+          trustMetaProfile
```
hossenlopp, line 99: I don't think this argument goes anywhere. It's not in the signature of `calculateDetailedResult`.

components/resource-creation/ResourcePanel.tsx
```diff
@@ -65,7 +65,7 @@ export default function ResourcePanel() {
                   measureBundle.content,
                   measurementPeriod.start?.toISOString(),
                   measurementPeriod.end?.toISOString(),
-                  useTrustMetaProfile
+                  trustMetaProfile
```
hossenlopp, line 68: Same comment as above.

pages/generate-test-cases.tsx
```diff
@@ -36,7 +36,7 @@ const TestCaseEditorPage: NextPage = () => {
               mb,
               start?.toISOString(),
               end?.toISOString(),
-              useTrustMetaProfile
+              trustMetaProfile
```
hossenlopp, line 39: Same issue as other situations that use this function.

hossenlopp (changes requested):
I think the option isn't getting passed into the calculation when it's being done for single patients.

pages/generate-test-cases.tsx
hossenlopp: Never mind! Github was showing me an out of date copy of the `calculateDetailedResult` function.

pages/index.tsx
```diff
@@ -86,7 +103,76 @@ const Home: NextPage = () => {
...
+                  label="Use trustMetaProfile"
+                  onLabel="YES"
+                  offLabel="NO"
+                  size="lg"
+                  checked={trustMetaProfile}
+                  onChange={event => setUseTrustMetaProfile(event.currentTarget.checked)}
+                ></Switch>
+              </Group>
+            </Grid.Col>
+          </Grid> */}
```
hossenlopp, line 175: Remove commented out stuff.

pages/index.tsx
```diff
@@ -28,10 +44,11 @@ const useStyles = createStyles(theme => ({
 
 const Home: NextPage = () => {
   const measureBundle = useRecoilValue(measureBundleState);
-
+  const [trustMetaProfile, setUseTrustMetaProfile] = useRecoilState(trustMetaProfileState);
   const [uploadSuccess, setUploadSuccess] = useState(false);
   const [errorLog, setErrorLog] = useState<MeasureUploadError[]>([]);
   const [datesValid, setDatesValid] = useState(false);
+  const [opened, setOpened] = useState(false);
```
hossenlopp, line 51: Suggest a more descriptive name. Maybe `trustMetaPopoverOpened`.

components/resource-creation/ResourceDisplay.tsx
hossenlopp: Never mind! Github was showing me an out of date copy of `calculateDetailedResult`.

components/resource-creation/ResourcePanel.tsx
hossenlopp: Never mind! Github was showing me an out of date copy of `calculateDetailedResult`.

hossenlopp (changes requested):
Just one naming suggestion and some commented out code to clean up.

pages/index.tsx
elsaperelli: LOL oops

hossenlopp (approved):
Looks good and works as expected!

PR #88, opened 2024-01-31
Multi group measures

elsaperelli:
# Summary
This PR adds support for multiple group measures in the MeasureHighlightingPanel.

## New Behavior
Before, fqm-testify only supported showing measure logic highlighting for one group of a patient; however, we have more recently come across multiple group measures such as CMS347 that has 4 groups. In the MeasureHighlightingPanel, there are tabs at the top for each of the groups (only 1 tab if it is not a multiple group measure). Each tab contains a PopulationComparisonTable and Measure highlighting for that patient and group.

## Code Changes
- `TestCaseEditor.tsx` - adds Mantine Tabs component for each group
- `MeasureHighlightingPanel.tsx` - updated the HTML to only return the HTML of a group on a patient, not all of the groups
- `PopulationComparisonTable.tsx` - adds state variable to the component so that it can be properly reused and takes in a detailed result (group result) rather than a patientId
- `PopulationComparisonTable.test.tsx` - updated unit test

# Testing Guidance
- `npm run check`
- `npm run dev` and navigate to `http://localhost:3000`
- Make sure that the app still works with single group measures (should have one tab for one group)
- Upload CMS347 make sure that the tabs work and that they display the correct results and highlighting for a patient and group


lmd59 (commented):
Looks good! My comments are just UI suggestions around the tab styling. 
It's a bit hard to tell that the tabs are tabs, maybe for a couple of reasons. Maybe we can make the text bold or some other styling to make it stand out from the other normal text on the page. Or maybe we could give the tabs an outline so that you can see the tab element itself before you hover. It may also be a bit confusing when nothing shows up before you select a tab. Maybe we can auto-select the first tab when the patient is first selected so that some content shows up for that action.

hossenlopp (changes requested):
Did a quick test of this. Works pretty well and minimally gets us to have the group selection in the highlighting output. Would be nice if it doesn't show the bar when there is only one group.

lmd59 (approved):

PR #97, opened 2024-12-29
Add option to remove resources not relevant to measure

elsaperelli:
# Summary
This PR adds an option on Patient Bundle upload to remove any resources on the Patient Bundle that are not relevant to the measure.

## New Behavior
There is now a switch on the Patient Bundle import modal that, when set, removes any resources on the Patient Bundle that do not match up with any of the Measure's data requirements. This allows the user to minimize the number of resources that are uploaded, while making sure that any necessary resources remain.

## Code Changes
- `README.md` - added sentence about removing irrelevant resources on import switch.
- `ImportModal.test.tsx` - added `mantineRecoilWrap` to the component now that it included a recoil state.
- `ImportModal.tsx` - added a Mantine switch for setting the minimize resources option and associated popover for further information.
- `PatientCreationPanel.tsx` - minimizes the resources on the test case that are uploaded if the switch is set to on.
- `resourceSwitch.ts` - atom to indicate if resources on a patient bundle are to be minimized using the data requirements of the measure.
- `dataRequirements.ts` - removed empty line in the imports
- `dataRequirementsByType.ts` - created a lookup selector object for the measure's data requirements where the key is the resource type and the attributes are `keepAll` (a boolean that defaults to false but is set to true if any of the data requirements do not have a codeFilter), `valueSets`, an array of value set urls, and `directCodes`, an array of Codings.
- `ValueSetHelper.ts` - contains two helper functions: one for getting codes and systems from a valueSet url, and one for taking a testCase, a measure bundle, and the data requirements by type lookup object and returning only the resources on the test case that are relevant to the data requirements.

# Testing Guidance
- `npm run check`
- `npm run dev`
- This is definitely a slightly challenging PR to test, as there are many different things that could go wrong. The positive is that since this is a switch, as long as the switch is set to off, the previously functionality of the app will remain.
- In order to test new functionality, a strategy I used was having this locally run and also pulling up the [testify site](https://projecttacoma.github.io/fqm-testify) and importing the same measure bundle and patient bundles and ensuring that clause coverage remains the same even if the irrelevant resources are removed.
- Try to break this! Test with a variety of different measures and test cases. The best way to see if any resources we need slipped through the cracks is to check clause coverage %. If clause coverage changes at all, that means that a resource got removed that should not have.


util/ValueSetHelper.ts
```diff
@@ -0,0 +1,196 @@
...
+                    const vsCodesAndSystems = getValueSetCodes(matchingDRType.valueSets, measureBundle);
+                    if (
+                      vsCodesAndSystems.find(vscas =>
+                        primaryCodeValue.coding?.find(c => c.code === vscas.code && c.system === vscas.system)
+                      )
+                    ) {
+                      newResources.push(r);
+                    }
+                  }
+                  if (matchingDRType.directCodes.length > 0) {
```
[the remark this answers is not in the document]
lmd59: This set of code from line 71 to 87 is also used a lot and might benefit from being pulled out into its own function.

util/ValueSetHelper.ts
```diff
@@ -0,0 +1,196 @@
...
+          } else if (primaryCodeInfo.codeType === 'FHIR.code') {
+            if (primaryCodeInfo.choiceType === true) {
+              if (primaryCodeInfo.multipleCardinality === true) {
+                // not sure if this happens based on codePaths.ts
+              } else {
+                // not sure if this happens based on codePaths.ts
+              }
+            } else {
+              if (primaryCodeInfo.multipleCardinality === true) {
+                // example: SearchParameter
```
lmd59, line 185: Are we planning on handling these cases?

lmd59 (commented):
There's some complex stuff happening here, so I really appreciate all of the commenting! Thanks for tackling a tricky feature!

I've left most comments inline and then have one more general user-focused comment:
Recommend using the `useRecoilState(resourceSwitchOn)` in the AppHeader to disable the measure edit button or give a warning when resources have been minimized. Or I believe we still have the original files loaded into memory, so we could re-trigger the minimization on re-calculation after uploading the new measure.

util/ValueSetHelper.ts
elsaperelli: Lol good question. I am really not sure and was having a hard time with these so I was wondering if you or @hossenlopp had any more insight if this is necessary (at least for now).

state/selectors/dataRequirementsByType.ts
```diff
@@ -0,0 +1,50 @@
...
+            }
+          }
+        } else {
+          if (dr.codeFilter === undefined) {
+            result[dr.type] = { keepAll: true, valueSets: [], directCodes: [] };
+          } else {
+            dr.codeFilter.forEach(cf => {
+              if (cf.valueSet) {
+                result[dr.type] = { keepAll: false, valueSets: [cf.valueSet], directCodes: [] };
+              } else if (cf.code) {
```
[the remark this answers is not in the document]
elsaperelli: Ah gotcha sounds good, I think I can just change this to two `if` statements like it's done in `util/fhir/codes.ts`

state/selectors/dataRequirementsByType.ts
```diff
@@ -0,0 +1,50 @@
...
+    const result: Record<string, DataRequirementsLookupByTypeProps> = {};
+
+    if (dataRequirements !== null) {
+      dataRequirements.forEach((dr, i) => {
+        if (result[dr.type]) {
+          if (result[dr.type].keepAll === false) {
+            if (dr.codeFilter === undefined) {
+              result[dr.type].keepAll = true;
+            } else {
+              dr.codeFilter.forEach(cf => {
```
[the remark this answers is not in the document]
elsaperelli: Hm I am a little confused by this comment. Since they are ANDed together, that sounds to me like they needed to be matched up to a resource exactly...but that sounds like it would cause a lot of issues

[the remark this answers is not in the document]
lmd59: Yeah- I think that's true if we were trying to be exact. But I think we're trying to do a helpful filter with this task rather than an exact filter? I think we're essentially treating them like an OR still instead of an AND, which means we'll still definitely keeping the relevant information, but we might have had the opportunity to filter down to even less information if we did the AND.

elsaperelli:
As a temporary and somewhat simple solution to your comments regarding making it clearer to the user when the resources on a Test Case have been minimized, I decided to add an attribute to the `TestCaseInfo` interface as a `minResources` boolean. This way the boolean is set to "true" if the minimize resources option was used on Import. I decided on this approach so that the user can decide to use this option on a Test Case by Test Case basis. This also makes it easier for the user to see which test cases were uploaded with this option even if they have also added test cases without that option. The only problem with this solution is it doesn't exactly address your suggestion of being able to run turn the option on and off when the test cases are already uploaded- I think that would be a little bit more complex and in the spirit of not scope creeping too much, I decided on this approach. I am still happy to make a follow up task to do that if that would be of interest in the future. For now, let me know if this makes sense!

state/selectors/dataRequirementsByType.ts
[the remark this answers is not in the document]
elsaperelli: Ah okay, that makes sense! Will add a comment and blub to the README!

util/ValueSetHelper.ts
```diff
@@ -0,0 +1,179 @@
...
+  const newResources: fhir4.BundleEntry[] = [];
+  testCase.resources.forEach(r => {
+    // throw out any resources that are not in any of the dataRequirements
+    // iterate over every resource in each bundle
+    // keep Patient and MeasureReport
+    if (
+      r.resource &&
+      r.resource?.resourceType &&
+      r.resource?.resourceType !== 'MeasureReport' &&
+      r.resource?.resourceType !== 'Patient'
```
hossenlopp, line 81: This logic may not be necessary. It looks like `bundleToTestCase` in `util/import.ts` already takes care of filtering them out of the testCase resource list.

util/ValueSetHelper.ts
hossenlopp: We can definitely skip over cases where nonsense resources are found. SearchParameter and OperationDefinition have no business in a patient bundle.

util/ValueSetHelper.ts
```diff
@@ -0,0 +1,179 @@
...
+}
+
+/**
+ * Helper function for repeated logic in minimizeTestCaseResources that takes a primaryCodeValue,
+ * a resource, a matching DataRequirement of the same resourceType, a measureBundle, and a newResources
+ * array, checks that the primaryCodeValue exists and if the primaryCodeValue matches any of the
+ * directCodes or ValueSet codes of a DataRequirement of the same resourceType, the resource is added
+ * to the newResources array
+ */
+function checkCodesAndValueSets(
```
hossenlopp, line 35: This function could be simplified to not deal with the `resource` and `newResources` at all. All this needs to do the code comparison and return a boolean if it matches or not. This would retain separation of concerns by only having the `newResources` be managed by the `minimizeTestCaseResources` function.

util/ValueSetHelper.ts
```diff
@@ -0,0 +1,179 @@
...
+  primaryCodeValue: fhir4.CodeableConcept,
+  matchingDRType: DataRequirementsLookupByTypeProps,
+  measureBundle: fhir4.Bundle | null,
+  resource: fhir4.BundleEntry,
+  newResources: fhir4.BundleEntry[]
+) {
+  if (primaryCodeValue) {
+    if (
+      matchingDRType.directCodes.length > 0 &&
+      matchingDRType.directCodes.find(dc => primaryCodeValue.coding?.find(c => c.code === dc.code))
```
hossenlopp, line 45:
```suggestion
      matchingDRType.directCodes.find(dc => primaryCodeValue.coding?.includes(c => c.code === dc.code))
```

`includes` should be used since we only need to know about existence in the array.

util/ValueSetHelper.ts
```diff
@@ -0,0 +1,179 @@
...
+                  } else if (matchingDRType.valueSets.length > 0) {
+                    const vsCodesAndSystems = getValueSetCodes(matchingDRType.valueSets, measureBundle);
+                    if (
+                      vsCodesAndSystems.find(
+                        vscas => primaryCodeValue.code === vscas.code && primaryCodeValue.system === vscas.system
+                      )
+                    ) {
+                      newResources.push(r);
+                    }
+                  }
```
hossenlopp, line 153: This looks like it is the same code in the `checkCodesAndValueSets` function.

util/ValueSetHelper.ts
```diff
@@ -0,0 +1,179 @@
...
+ * an array of BundleEntries that are TestCase resources that are relevant to the
+ * measure bundle's data requirements
+ */
+export function minimizeTestCaseResources(
+  testCase: TestCaseInfo,
+  measureBundle: fhir4.Bundle | null,
+  drLookupByType: Record<string, DataRequirementsLookupByTypeProps>
+): fhir4.BundleEntry[] {
+  const newResources: fhir4.BundleEntry[] = [];
+  testCase.resources.forEach(r => {
```
hossenlopp, line 73: There are multiple places that `newResources.push` is called here. It feels like this could have a chance for duplicated resources, likely when there are valueSets with overlapping codes. This could be changed to using `Array.filter` with this predicate function returning `true` when ever matching code is found. This would also allow for an easy short-circuit to help speed things up.

hossenlopp (changes requested):
The JIRA task stated we should just focus on the primaryCodePath of each resource. The logic is set up well for doing this in the `minimizeTestCaseResources` function. But the logic that takes the dataRequirements and creates the nice mapping grabs all of the codeFilters on the dataRequirements, even if they are for paths other than the primaryCodePath. This could result in codes/valuesets from a codeFilter that is NOT on the primaryCodePath to later be used in the comparison against the primaryCodePath.

I think we could keep operating with primaryCodePath limitation and sort out the dataRequirements lookup function to filter the codeFilters it finds.

But it may be fairy possible to make this work with ALL paths. It would require the map of dataRequirements to have a second level mapping for each path to the possible codes or valuesets. This would also require a few more of the funny cases with the weird cardinalities to be handled in the minimizeTestCaseResources.

There also a few suggestions I have on keeping separation of concerns and avoiding duplicates.

util/ValueSetHelper.ts
elsaperelli: If I make this change, then I get a type error with the `.find` that it is in, should I change them both?

util/ValueSetHelper.ts
elsaperelli: It's a little different

util/ValueSetHelper.ts
elsaperelli: I am actually not totally seeing how resources could be duplicated here, can you point to a spot in the logic where that could happen?

util/ValueSetHelper.ts
hossenlopp: Yep. Both would need to be `.includes`. and both predicates should be returning booleans.

util/ValueSetHelper.ts
```diff
@@ -0,0 +1,182 @@
...
+                  r.resource,
+                  codeInfo.primaryCodePath
+                ) as fhir4.CodeableConcept[];
+                if (primaryCodeValue) {
+                  primaryCodeValue.forEach(pcv => {
+                    const matchingCode = checkCodesAndValueSets(pcv, matchingDRType, measureBundle);
+                    if (matchingCode) {
+                      newResources.push(r);
+                    }
+                  });
```
hossenlopp, line 116: This is one of the places were we could end up with duplicated resources. Since the field here is a CodeableConcept[] there effectively is a list of of list of codes. There could be a situation where there are multiple codes being used to describe the same thing but using different code systems for compatibility. Many eCQM ValueSets are grouping ValueSets that have multiple code systems for representing the same thing for compatibility. So it is very likely that an Encounter could have two codes that are both in the ValueSet for a data requirement. So this logic would add that Encounter to the newResources array twice.

Switching the entire `forEach` loop over the `testCase.resources` to a `filter` and changing the logic to `return true` when we find a match would completely eliminate the chance for duplicated resources to occur. It would also us to move on to the next resource immediately when we have found one we want to keep.

util/ValueSetHelper.ts
elsaperelli: image

lmd59:
> As a temporary and somewhat simple solution to your comments regarding making it clearer to the user when the resources on a Test Case have been minimized, I decided to add an attribute to the `TestCaseInfo` interface as a `minResources` boolean. This way the boolean is set to "true" if the minimize resources option was used on Import. I decided on this approach so that the user can decide to use this option on a Test Case by Test Case basis. This also makes it easier for the user to see which test cases were uploaded with this option even if they have also added test cases without that option. The only problem with this solution is it doesn't exactly address your suggestion of being able to run turn the option on and off when the test cases are already uploaded- I think that would be a little bit more complex and in the spirit of not scope creeping too much, I decided on this approach. I am still happy to make a follow up task to do that if that would be of interest in the future. For now, let me know if this makes sense!
> 
> Addressed in: [a930691](https://github.com/projecttacoma/fqm-testify/commit/a930691be52e8b833c0d1722c0b92a18e2b51d2e)

I think this was definitely a good add, and I agree that applying the slimming after the patient is uploaded is out of scope. I still think it would be nice to put some sort of warning on the edit bundle button at the top so that the user knows that the test cases that don't have all of their original resources may not calculate properly with a new measure.
Screenshot 2025-01-16 at 2 13 45 PM
We could maybe do a warning modal if the user clicks this button while there are any test cases with the minResources boolean.

util/ValueSetHelper.ts
hossenlopp:
```suggestion
      matchingDRType.directCodes.some(dc => primaryCodeValue.coding?.some(c => c.code === dc.code))
```

[`some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) is actually what i was looking for this case.

hossenlopp (approved):

lmd59 (approved):
Nice! Thanks for all of the work on making this come together!

hossenlopp (approved):

PR #98, opened 2025-01-16
1582 Show Date on Resource

zacharyrobin:
# Summary
Added a date field to the resource card.  [Jira Ticket 1582](https://jira.mitre.org/browse/TACOSTRAT-1582)

## New Behavior
Resource card will display either the dateTime, date, or period depending on what the resource has populated. Additionally has a tooltip on hover displaying the date type.
Screenshot 2025-01-16 at 7 30 43 AM

## Code Changes
Changes are primarily to `ResourceDisplay.tsx` and a small change in `ResourceinfoCard.tsx`. 

Notes:
- If desired I can change the way the date is displayed easily.  
- Because of the multiple formats and the inconsistent naming convention, the code to display it was a little more convoluted than Id like.  Open to suggestions if there is a streamlined way to do this. 

# Testing Guidance
start up `fqm-testify` and upload any bundle and a test.

### DIFF
--- a/components/resource-creation/ResourceDisplay.tsx
+++ b/components/resource-creation/ResourceDisplay.tsx
@@ -87,10 +87,11 @@ function ResourceDisplay() {
     setSelectedDataRequirement({ name: '', content: null });
   };
 
-  const dateForResource = (resource: fhir4.FhirResource) => {
+  // Return type is a an string array, the first element is the formatted date and the second is the resource type.
+  const dateForResource = (resource: fhir4.FhirResource): string[] => {
     const dateInfo = PrimaryDatePaths.parsedPrimaryDatePaths[resource.resourceType];
     if (!resource || !PrimaryDatePaths?.parsedPrimaryDatePaths || !dateInfo) {
-      return 'N/A';
+      return ['N/A', 'N/A'];
     }
 
     for (const nameOfResourceDate of Object.keys(dateInfo)) {
@@ -103,18 +104,7 @@ function ResourceDisplay() {
         }
         // If the only dataType is either dateTime or date
         else {
-          return (
-            <Tooltip
-              arrowPosition="side"
-              arrowOffset={25}
-              arrowSize={8}
-              label={nameOfResourceDate}
-              withArrow
-              position="top-start"
-            >
-              <Text>{formatDate(fhirpath.evaluate(resource, nameOfResourceDate)[0])}</Text>
-            </Tooltip>
-          );
+          return [formatDate(fhirpath.evaluate(resource, nameOfResourceDate)[0]), nameOfResourceDate];
         }
       }
 
@@ -129,37 +119,20 @@ function ResourceDisplay() {
             }
             // Else if dataType === 'dateTime' || 'Date'
             else {
-              return (
-                <Tooltip
-                  arrowPosition="side"
-                  arrowOffset={25}
-                  arrowSize={8}
-                  label={fullResourceDateName}
-                  withArrow
-                  position="top-start"
-                >
-                  <Text>{formatDate(fhirpath.evaluate(resource, fullResourceDateName)[0])}</Text>
-                </Tooltip>
-              );
+              return [formatDate(fhirpath.evaluate(resource, fullResourceDateName)[0]), fullResourceDateName];
             }
           }
         }
       }
     }
-    return 'N/A';
+    return ['N/A', 'N/A'];
   };
 
   // Formatter for if the date is a period
   const formatPeriod = (resource: fhir4.FhirResource, resourcePeriod: string, dateTypeName: string) => {
     const startTime = fhirpath.evaluate(resource, resourcePeriod + '.start')[0];
     const endTime = fhirpath.evaluate(resource, resourcePeriod + '.end')[0];
-    return (
-      <Tooltip arrowPosition="side" arrowOffset={25} arrowSize={8} label={dateTypeName} withArrow position="top-start">
-        <Text>
-          {formatDate(startTime)} - {formatDate(endTime)}
-        </Text>
-      </Tooltip>
-    );
+    return [formatDate(startTime) + '-' + formatDate(endTime), dateTypeName];
   };
 
   // Using date-fns to format (other date formats available)
--- a/components/resource-creation/ResourceDisplay.tsx
+++ b/components/resource-creation/ResourceDisplay.tsx
@@ -1,4 +1,4 @@
-import { Stack } from '@mantine/core';
+import { Stack, Container, Text, Tooltip } from '@mantine/core';
 import { useCallback, useEffect, useState } from 'react';
 import produce from 'immer';
 import { useRecoilState, useRecoilValue, useSetRecoilState } from 'recoil';
@@ -20,6 +20,8 @@ import { detailedResultLookupState } from '../../state/atoms/detailedResultLooku
 import { DetailedResult } from '../../util/types';
 import { calculateDetailedResult } from '../../util/MeasureCalculation';
 import { trustMetaProfileState } from '../../state/atoms/trustMetaProfile';
+import { PrimaryDatePaths } from 'fhir-spec-tools'; 
+import { format } from 'date-fns';
 
 function ResourceDisplay() {
   const [currentTestCases, setCurrentTestCases] = useRecoilState(patientTestCaseState);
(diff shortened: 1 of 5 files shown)


components/resource-creation/ResourceDisplay.tsx
```diff
@@ -84,6 +86,70 @@ function ResourceDisplay() {
...
+        for (const dataType of dateinfo[nameOfResourceDate].dataTypes) {
+          // Capitalize the first char of dataType and append for proper resource date name
+          const fullResourceDateName = nameOfResourceDate + dataType.charAt(0).toUpperCase() + dataType.slice(1);
+          if (resource[fullResourceDateName]) {
+            if (dataType == 'Period') {
+              return formatPeriod(resource[fullResourceDateName], fullResourceDateName);
+            }
+            // Else if dataType == 'dateTime' || 'Date'
+            return (
+              <Tooltip arrowPosition='side' arrowOffset={25} arrowSize={8} label={'Date Type: ' + fullResourceDateName} withArrow position='top-start'>
```
hossenlopp, line 127:
```suggestion
              <Tooltip arrowPosition='side' arrowOffset={25} arrowSize={8} label={fullResourceDateName} withArrow position='top-start'>
```
I think just the name of the field we got the date from is enough info here. Neat feature though!

components/resource-creation/ResourceDisplay.tsx
```diff
@@ -84,6 +86,70 @@ function ResourceDisplay() {
...
+            )
+          }
+        }
+      }
+    }
+    return 'No Date Found'
+  }
+
+  // Formatter for if the date is a period
+  const formatPeriod = (resourcePeriod: any, dateTypeName: string) => {
```
hossenlopp, line 139:
```suggestion
  const formatPeriod = (resourcePeriod: fhir4.Period, dateTypeName: string) => {
```
You can make use of the `fhir4` TS types here. Using these will also show you that you'll need to handle the fact that `start` and `end` may not be defined to note a period that doesn't end (ex AllergyIntollerance) or have a known start time.

components/resource-creation/ResourceDisplay.tsx
```diff
@@ -84,6 +86,70 @@ function ResourceDisplay() {
...
+      }
+      
+      // If isChoiceType is true
+      else {
+        for (const dataType of dateinfo[nameOfResourceDate].dataTypes) {
+          // Capitalize the first char of dataType and append for proper resource date name
+          const fullResourceDateName = nameOfResourceDate + dataType.charAt(0).toUpperCase() + dataType.slice(1);
+          if (resource[fullResourceDateName]) {
+            if (dataType == 'Period') {
+              return formatPeriod(resource[fullResourceDateName], fullResourceDateName);
```
hossenlopp, line 123:
```suggestion
              return formatPeriod(resource[fullResourceDateName] as fhir4.Period, fullResourceDateName);
```
Casting to the type now that we know it before passing it to the function. See comment on `formatPeriod` for the other end.

hossenlopp (changes requested):
Quick review. This is a neat feature. Got some suggestions for clearing up the `any` around the `Period`.

I also suggest letting this fail quietly when we cannot figure out any date information instead of throwing errors.

elsaperelli:
Looks like the checks aren't passing because you added an attribute to `ResourceInfoCardProps` (`date`) and it also needs to be added to the corresponding `MOCK_CALLBACK_PROPS` in `ResourceInfoCardProps.test.tsx`. I think following what is done there for `resourceType` and `label` will be useful, but let me know if you need help!

components/resource-creation/ResourceDisplay.tsx
```diff
@@ -84,6 +87,79 @@ function ResourceDisplay() {
...
+          }
+        }
+      }
+    }
+    return 'N/A'
+  }
+
+
+  // Formatter for if the date is a period
+  const formatPeriod = (resource: fhir4.Period, resourcePeriod: string, dateTypeName: string) => {
```
elsaperelli, line 144: Also it looks like there is some mismatch with variable names and types here...I think that the resource name of the fhir4.Period is potentially misleading, I would change that to period: fhir4.Period or something.

elsaperelli: On further investigation, I think resource should be of type fhir4.FhirResource

components/resource-creation/ResourceDisplay.tsx
```diff
@@ -84,6 +87,79 @@ function ResourceDisplay() {
...
+      return 'N/A'
+    }
+
+    for (const nameOfResourceDate of Object.keys(dateinfo)) {
+      // If only one dataType
+      const resourceDateData = fhirpath.evaluate(resource, nameOfResourceDate)[0]
+      if(dateinfo[nameOfResourceDate].dataTypes.length == 1 && resourceDateData){
+        // If the only dataType is a period 
+        if(dateinfo[nameOfResourceDate].dataTypes[0] == 'Period'){
+          return formatPeriod(resource as fhir4.Period, nameOfResourceDate, nameOfResourceDate)
```
elsaperelli, line 106:
```suggestion
          return formatPeriod(resource, nameOfResourceDate, nameOfResourceDate)
```
We don't need to cast this fhir4.FhirResource type to a fhir4.Period, just leave it as a fhir4.FhirResource and change the input type of resource in `formatPeriod`

components/resource-creation/ResourceDisplay.tsx
```diff
@@ -84,6 +87,79 @@ function ResourceDisplay() {
...
+      }
+      
+      // If isChoiceType is true
+      else {
+        for (const dataType of dateinfo[nameOfResourceDate].dataTypes) {
+          // Capitalize the first char of dataType and append for proper resource date name
+          const fullResourceDateName = nameOfResourceDate + dataType.charAt(0).toUpperCase() + dataType.slice(1);
+          if (fhirpath.evaluate(resource, fullResourceDateName)[0]) {
+            if (dataType == 'Period') {
+              return formatPeriod(resource as fhir4.Period, fullResourceDateName, fullResourceDateName);
```
elsaperelli, line 125:
```suggestion
              return formatPeriod(resource, fullResourceDateName, fullResourceDateName);
```
See above comment

elsaperelli (changes requested):
Great stuff Zach!!!

Thanks for changing your code to use `fhirpath`, I know that was probably annoying and I think a good chunk of my suggestions are just left over from that transition, but I think the fhirpath use allows us to be more specific with typing!

components/resource-creation/ResourceDisplay.tsx
```diff
@@ -84,6 +87,79 @@ function ResourceDisplay() {
...
 
+  const dateForResource = (resource: fhir4.FhirResource ) => {
+    if (!resource || !PrimaryDatePaths?.parsedPrimaryDatePaths) {
+      return 'N/A'
+    }
+
+    const dateinfo =  PrimaryDatePaths.parsedPrimaryDatePaths[resource.resourceType] 
+    if (!dateinfo) {
+      return 'N/A'
+    }
```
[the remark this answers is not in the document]
zacharyrobin: yea initially it was two different error messages but I removed them so definitely should combine them

components/resource-creation/ResourceDisplay.tsx
```diff
@@ -84,6 +87,79 @@ function ResourceDisplay() {
...
+          if (fhirpath.evaluate(resource, fullResourceDateName)[0]) {
+            if (dataType == 'Period') {
+              return formatPeriod(resource as fhir4.Period, fullResourceDateName, fullResourceDateName);
+            }
+            // Else if dataType == 'dateTime' || 'Date'
+            else {
+              return (
+                <Tooltip arrowPosition='side' arrowOffset={25} arrowSize={8} label={fullResourceDateName} withArrow position='top-start'>
+                  <Text>{formatDate(fhirpath.evaluate(resource, fullResourceDateName)[0])}</Text>
+                </Tooltip>
```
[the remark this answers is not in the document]
zacharyrobin: thats a good point, yea might be easier to understand and less redundant if I combine / sync the return type

components/resource-creation/ResourceDisplay.tsx
```diff
@@ -129,37 +119,20 @@ function ResourceDisplay() {
...
-                  arrowOffset={25}
-                  arrowSize={8}
-                  label={fullResourceDateName}
-                  withArrow
-                  position="top-start"
-                >
-                  <Text>{formatDate(fhirpath.evaluate(resource, fullResourceDateName)[0])}</Text>
-                </Tooltip>
-              );
+              return [formatDate(fhirpath.evaluate(resource, fullResourceDateName)[0]), fullResourceDateName];
```
elsaperelli, line 122: 'Text' and 'Tooltip' can be removed from the top (see lint error warning when running `npm run check`)

components/utils/ResourceInfoCard.tsx
```diff
@@ -4,7 +4,7 @@ import { Edit, Trash } from 'tabler-icons-react';
 export interface ResourceInfoCardProps {
   resourceType: string;
   label: string;
-  date: string | React.ReactElement;
+  date: string[];
```
elsaperelli, line 7: Rather than making this an Array of string where we know it will only have two values and the first and second both correspond with a specific value, I would make use of this interface. I would just make date a string and then add another attribute to this interface for `datePath` or something like that.

components/resource-creation/ResourceDisplay.tsx
```diff
@@ -84,6 +87,63 @@ function ResourceDisplay() {
     setSelectedDataRequirement({ name: '', content: null });
   };
 
+  // Return type is a an string array, the first element is the formatted date and the second is the resource type.
```
elsaperelli, line 90:
```suggestion
  // Return type is a string array, the first element is the formatted date and the second is the resource type.
```
But I would change this type anyways (see below comment)

elsaperelli (changes requested):
Getting close!! Just one more main suggestion :)

elsaperelli (approved):
Woooo great job!!

PR #100, opened 2025-02-26
Use fhir-spec-tools in place of previous scripts

elsaperelli:
# Summary
This PR adds [fhir-spec-tools](https://github.com/projecttacoma/fhir-spec-tools) as a dependency in order to replace previous scripts and script outputs that fhir-spec-tools was intended to replace. 

## New behavior
There should be no new behavior

## Code changes
- `package.json` / `package-lock.json` - add `fhir-spec-tools` as a dependency, remove "parse-compartment-def" command
- `__tests__/utils/fhir/dates.test.ts` -  replaced `parsePrimaryDatePath` with fhir-spec-tools, but fhir-spec-tools does not export `DateFieldInfo`, so I just put it in `util/fhir/dates.ts`
- `package.json` - removed commands no longer used
- `pages/api/hello.ts` - I don't think we ever used this and that it was just leftover from creating the repo so I deleted it lol
- `scripts/parseCodePath.ts` / `scripts/parsePrimaryDatePath.ts` - scripts replaced with fhir-spec-tools
- `util/codePaths.ts` / `util/referencesMap.ts` / `primaryDatePaths.ts` - output file replaced with fhir-spec-tools
- `state/selectors/dataRequirementsLookupByType.ts` / `util/ValueSetHelper.ts` / `util/fhir/codes.ts` - replace parsedCodePaths reference
- `util/fhir/patient.ts` - replace patient references map with `patientAttributePaths` from fhir-spec-tools
- `util/fhir/dates.ts` - replace parsedPrimaryDatePaths reference, added interfaces since they are not exported from fhir-spec-tools

# Testing guidance
- `npm run check`
- `npm run dev`
- Go through the app and make sure the functionality is all the same and that nothing breaks

## Question
- I tried deleting `fixtures/model-info/fhir-modelinfo-4.0.1.xml` since I didn't think it was needed anymore, but I got this weird warning when I did `npm run dev`: 
- 
image


zacharyrobin (commented):
I'm not seeing the same error when I remove `fixtures/model-info/fhir-modelinfo-4.0.1.xml`

zacharyrobin (approved):
Ran through some tests and functionality works as expected, looks good to me.

PR #101, opened 2025-03-28
Search/Sort of a Patients' Resources

zacharyrobin:
# Summary
Adding functionality to `fqm-testify` to allow users to search and sort through the resources. 

*Note that this is a draft
TODO:
1. I found when testing with larger data sets there was some lag. I think this may be because `getFhirResourceSummary()` and `dateForResource()` are called every time there is a search or sort? If this is the reason why then a potential solution may be to add an interface like the following so that the card data is mapped and set once: 
```
export interface Resource {
  key: Key | null | undefined;
   resourceType: string;
   label: string;
   date: DateInfo;
 }


// and then have some function to set them 

  const resources: Resource[] = [];

    for (const entry of entries) {
      if (!entry.resource) continue;

      const resource: Resource = {
        key: entry.resource.id ?? null,
        resourceType: entry.resource.resourceType,
        label: getFhirResourceSummary(entry.resource),
        date: dateForResource(entry.resource),
      };

      resources.push(resource);
    }

    return resources;
  }
```
I dont know if there is a better way to do this 

2. I believe the current `ResourceDisplay.test.tsx` may be broken as it just continuously runs, Im unsure of how I should go about solving this. 
3. I currently have the search set to search for the `date`, `resourceType`, and `resourceSummary` but can remove `resourceSummary` if desired. 

## New Behavior
- Added search bar to `Resource Display` view with ability to search by `date`, `resourceType`, and `resourceSummary`
- Added buttons to sort by `date` or `resourceType`
Screenshot 2025-03-27 at 11 54 18 PM

## Code Changes
- Created ``ResourceSearchSort.tsx` to handle the filter functions including a search bar and sort buttons
- Added `ResourceSearchSort` to `ResourceDisplay.tsx`
- Modified `ResourceDisplay.tsx` to update resource list when ResourceSearchSort has a change
- Added test for `ResourceDisplay.tsx` under ``ResourceDisplay.test.tsx`

# Testing Guidance
- Start up `fqm-testify` and upload any bundle and a test
- Check that the sort buttons work for ascending and descending order 
- Check that search function works as expected

### DIFF
--- a/components/resource-creation/ResourceDisplay.tsx
+++ b/components/resource-creation/ResourceDisplay.tsx
@@ -320,7 +320,6 @@ function ResourceDisplay() {
   //   return resources;
   // }
 
-
   // Handles changes from ResourceSearchSort
   const handleSrotedResources = (sorted: fhir4.BundleEntry[]) => {
     setSortedResources(sorted);
@@ -344,7 +343,11 @@ function ResourceDisplay() {
       {selectedPatient && selectedDataRequirement && currentTestCases[selectedPatient].resources.length > 0 && (
         <>
           {/* Passing in the selected patient resources to allow for searching + sorting */}
-          <ResourceSearchSort resources={currentTestCases[selectedPatient].resources} onSorted={handleSrotedResources} dateForResource={dateForResource} />
+          <ResourceSearchSort
+            resources={currentTestCases[selectedPatient].resources}
+            onSorted={handleSrotedResources}
+            dateForResource={dateForResource}
+          />
 
           <Stack data-testid="resource-display-stack">
             {sortedResources.map(bundleEntry => {
--- a/components/resource-creation/ResourceDisplay.tsx
+++ b/components/resource-creation/ResourceDisplay.tsx
@@ -1,28 +1,36 @@
 import { Stack } from '@mantine/core';
-import { useCallback, useEffect, useState } from 'react';
-import produce from 'immer';
-import { useRecoilState, useRecoilValue, useSetRecoilState } from 'recoil';
-import CodeEditorModal from '../modals/CodeEditorModal';
-import { measureBundleState } from '../../state/atoms/measureBundle';
-import { selectedDataRequirementState } from '../../state/atoms/selectedDataRequirement';
-import { patientTestCaseState, TestCase } from '../../state/atoms/patientTestCase';
-import { selectedPatientState } from '../../state/atoms/selectedPatient';
-import { measurementPeriodState } from '../../state/atoms/measurementPeriod';
-import ConfirmationModal from '../modals/ConfirmationModal';
-import ResourceInfoCard from '../utils/ResourceInfoCard';
-import { calculationLoading } from '../../state/atoms/calculationLoading';
 import { showNotification } from '@mantine/notifications';
 import { IconAlertCircle } from '@tabler/icons';
-import { WritableDraft } from 'immer/dist/internal';
-import { createFHIRResourceString } from '../../util/fhir/resourceCreation';
-import { getFhirResourceSummary } from '../../util/fhir/codes';
-import { detailedResultLookupState } from '../../state/atoms/detailedResultLookup';
-import { DetailedResult } from '../../util/types';
-import { calculateDetailedResult } from '../../util/MeasureCalculation';
-import { trustMetaProfileState } from '../../state/atoms/trustMetaProfile';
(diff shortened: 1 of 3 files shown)


components/resource-creation/ResourceSearchSort.tsx
```diff
@@ -0,0 +1,98 @@
...
+          const typeB = b.resource.resourceType.toLowerCase();
+          return sortOrder === 'asc' ? typeA.localeCompare(typeB) : typeB.localeCompare(typeA);
+        }
+
+        return 0;
+      });
+    }
+
+    // checking for a change in the results to prevent infinite loop
+    const newHash = filtered.map(f => f.resource?.id || '').join(',');
```
lmd59, line 71: I know that this is a fix to prevent an infinite loop on loading this state, but it seems like a not really REACT-y way of managing a state change. It does work, but I'm waffling on whether this is a good way to handle this situation. I think that @hossenlopp said he was interested in taking a look at this PR, so I might loop him in here to get his thoughts.

components/resource-creation/ResourceSearchSort.tsx
```diff
@@ -0,0 +1,98 @@
...
+    if (prevResultRef.current !== newHash) {
+      prevResultRef.current = newHash;
+      onSorted(filtered);
+    }
```
lmd59, line 75: I know that this is a fix to prevent an infinite loop on loading this state, but it seems like a not really REACT-y way of managing a state change. It does work, but I'm waffling on whether this is a good way to handle this situation. I think that @hossenlopp said he was interested in taking a look at this PR, so I might loop him in here to get his thoughts.

zacharyrobin: I know its not pretty, open to alternatives. I found it to be the simplest but I can also rework some of the architecture as we discussed last friday if thats better.

components/resource-creation/ResourceDisplay.tsx
```diff
@@ -292,9 +221,51 @@ function ResourceDisplay() {
     }
   };
 
-  const handleSortedResources = (sorted: fhir4.BundleEntry[]) => {
-    setSortedResources(sorted);
-  };
+  // Card Filtering
+  useEffect(() => {
+    if (selectedPatient && selectedDataRequirement && currentTestCases[selectedPatient].resources.length > 0) {
```
lmd59, line 226: Do we need to check `selectedDataRequirement`?

components/resource-creation/ResourceDisplay.tsx
```diff
@@ -292,9 +221,51 @@ function ResourceDisplay() {
...
+          return 0;
+        });
+      }
+
+      setFilteredPatientResources(filtered);
+    }
+  }, [
+    cardFilters.searchString,
+    cardFilters.sortType,
+    cardFilters.sortOrder,
```
lmd59, line 264: Recommend simplifying these 3 to just `cardFilters` rather than calling out each field individually.

components/resource-creation/ResourceDisplay.tsx
```diff
@@ -292,9 +221,51 @@ function ResourceDisplay() {
...
+    selectedPatient,
+    currentTestCases,
+    selectedDataRequirement
```
lmd59, line 267: Recommend remove `selectedDataRequirement` if it's not needed.

components/resource-creation/ResourceSearchSort.tsx
```diff
@@ -1,87 +1,47 @@
...
+import { cardFiltersAtom } from '../../state/atoms/patientResources';
 
-type Props = {
-  resources: fhir4.BundleEntry[];
-  onSorted: (sortedResources: fhir4.BundleEntry[]) => void;
-  dateForResource: (resource: fhir4.FhirResource) => { date: string; dateType: string };
-};
+const ResourceSearchSort = () => {
+  const setCardFilters = useSetRecoilState(cardFiltersAtom);
+  const cardFilters = useRecoilValue(cardFiltersAtom);
```
lmd59, line 9:
```suggestion
  const [cardFilters, setCardFilters] = useRecoilState(cardFiltersAtom);
```
You can simplify these to one line (will also need to update recoil imports).

lmd59 (commented):
Looking good! A couple of small comments that I think should be quick!

lmd59 (approved):
Nice!

PR #102, opened 2025-04-09
Add a Delete All Patients button

elsaperelli:
# Summary
This PR adds a "Delete All Patients" button in the PatientCreationPanel.

## New Behavior
The user can now delete all patients in the PatientCreationPanel with one button rather than going through each patient.

## Code Changes
- `PatientCreationPanel.tsx` - added `isDeleteAll` state variable so that the Confirmation Modal text can be easily adjusted for the new button (is there a better way to do this so I can still reuse that Confirmation Modal?), add `deleteAllPatients` to remove all patients and their resources
- `PatientCreationPanel.test.tsx` - associated unit test

# Testing Guidance
- `npm run check`
- `npm run dev`
- Make sure the app has the desired behavior
- Looking for feedback on the UI and button placement- see slack thread for details!

### DIFF
--- a/components/patient-creation/PatientCreationPanel.tsx
+++ b/components/patient-creation/PatientCreationPanel.tsx
@@ -49,7 +49,7 @@ function PatientCreationPanel() {
   const [copiedPatient, setCopiedPatient] = useState<string | null>(null);
   const [isConfirmationModalOpen, setIsConfirmationModalOpen] = useState(false);
   const [isImportModalOpen, setIsImportModalOpen] = useState(false);
-  const [isDeleteAll, setIsDeleteAll] = useState(false);
+  const [patientsToDelete, setPatientsToDelete] = useState<string[]>([]);
   const [currentPatients, setCurrentPatients] = useRecoilState(patientTestCaseState);
   const currentTestMRLookup = useRecoilValue(cqfmTestMRLookupState);
   const [selectedPatient, setSelectedPatient] = useRecoilState(selectedPatientState);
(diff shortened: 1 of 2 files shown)


components/patient-creation/PatientCreationPanel.tsx
```diff
@@ -237,6 +249,9 @@ function PatientCreationPanel() {
 
   const getConfirmationModalText = (patientId: string | null) => {
     let patientName;
+    if (isDeleteAll) {
```
lmd59, line 252: This is a bit futzy since this function acts on a passed patient name, which for delete all is irrelevant. If we want to lean into the `isDeleteAll` state, then I would have this function be reliant on a passed boolean for deleting all... but there might be a better way.

lmd59: My two potential recommended solutions:
A. Create another instance of the `ConfirmationModal` which is configured specifically for deleting all patients. In this case the philosophy would be to have one modal for each button we'd potentially press.
B. Configure our current instance of the `ConfirmationModal` based on the list of patients we want to delete. We could replace the `isDeleteAll` state with something like `patientsToDelete` state, containing an array of all patient ids that we want to delete. For a single patient delete, we'd set this state to [`selectedPatient`], and for all patient delete, we'd set this state to `currentPatients`. Then we should be able to configure the ConfirmationModal based on this state, potentially with a message like `Deleting # patients`. This philosophy is that we're creating this instance of the Confirmation Modal to handle "deleting patients" in whatever number. This would set us up to more easily extend the logic to deleting some checked selection of patients (which is a future task in our backlog).

lmd59 (commented):
Nice! Definitely works. A couple of thoughts for some potential alternatives for encapsulation and state organization.

lmd59 (commented):
A couple of small changes recommended in line. Looking good!

components/patient-creation/PatientCreationPanel.tsx
```diff
@@ -441,6 +456,19 @@ function PatientCreationPanel() {
           <IconFileDownload />
           &nbsp;Download All
         </Button>
+        <Button
+          aria-label="Delete All Patients"
+          disabled={Object.keys(currentPatients).length === 0}
+          onClick={() => {
+            setPatientsToDelete(Object.keys(currentPatients).map(cp => cp));
```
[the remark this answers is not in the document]
elsaperelli: LOL

lmd59 (approved):
Nice!

PR #104, opened 2025-04-15
Use statement-level HTML for Measure Highlighting

elsaperelli:
# Summary
This PR finally updates `fqm-testify` to use `buildStatementLevelHTML: true` in CalculationOptions so we can use `statementResult.statementLevelHTML` for the MeasureHighlightingPanel rather than parsing through the entire `DetailedPopulationGroupResult.html`.

## New Behavior
- There should be no new behavior observed in the app.

## Code Changes
- `components/calculation/MeasureHighlightingPanel.tsx` - `parsedHTML` now goes through the dr.statementResults, filters only the relevantStatements, and sorts them (this was all done in the regular `generateHTML` function in `fqm-execution` so this was all already done before when we parsed the overall html.
- `util/MeasureCalculation.ts` - add `buildStatementLevelHTML: true` to `calculateDetailedResult` calculationOptions
- `util/MeasurePopulations.ts` - added `sortStatements()` function (see question about this below

# Testing Guidance
- `npm run check`
- `npm run dev`
- Make sure the `MeasureHighlightingPanel.tsx` has the same functionality- CQL statement search still works, the statement result is still in the drop down below each statement, population groupId is still displayed, etc.

# Questions
- I had to pull that `sortStatements` function from fqm-execution, should I make a backlog task to perhaps have that function exported from fqm-execution so we don't need to add the whole function to this repository?

### DIFF
--- a/components/calculation/MeasureHighlightingPanel.tsx
+++ b/components/calculation/MeasureHighlightingPanel.tsx
@@ -34,10 +34,10 @@ export default function MeasureHighlightingPanel({ dr }: MeasureHighlightingPane
     return relevantStatements.map(sr => {
       if (sr.statementLevelHTML) {
         return (
-          <>
+          <div key={`${dr.groupId}-${sr.libraryName}-${sr.statementName}`}>
             {parse(sr.statementLevelHTML)}
             <PrettyOutput statement={sr} />
-          </>
+          </div>
         );
       }
     });
--- a/components/calculation/MeasureHighlightingPanel.tsx
+++ b/components/calculation/MeasureHighlightingPanel.tsx
@@ -1,10 +1,12 @@
 import { ActionIcon, Autocomplete, ScrollArea, Space, Text, createStyles } from '@mantine/core';
-import parse, { domToReact } from 'html-react-parser';
-import { useState } from 'react';
-import { Element as DomElement } from 'domhandler';
+import parse from 'html-react-parser';
+import { useMemo, useState } from 'react';
 import { Search, X } from 'tabler-icons-react';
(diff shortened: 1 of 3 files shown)


components/calculation/MeasureHighlightingPanel.tsx
```diff
@@ -21,24 +23,25 @@ export interface MeasureHighlightingPanelProps {
...
+    const relevantStatements = dr.statementResults.filter(s => s.relevance !== Relevance.NA);
+    sortStatements(measure, dr.groupId, relevantStatements);
+    return relevantStatements.map(sr => {
+      if (sr.statementLevelHTML) {
         return (
           <>
-            {domToReact([elem])}
-            <PrettyOutput statement={statementResult} />
+            {parse(sr.statementLevelHTML)}
+            <PrettyOutput statement={sr} />
```
zacharyrobin, line 39: It looks like I'm getting a unique key error 
Screenshot 2025-04-16 at 7 42 31 AM
that is from here. wrapping it with some unique key should fix it something like:
``` 
const key = `${dr.groupId}-${sr.libraryName}-${sr.statementName}`;
<div key={key}>
        {parse(sr.statementLevelHTML)}
        <PrettyOutput statement={sr} />
      </div>

zacharyrobin (changes requested):
Looks great! just one small change to fix a warning.

components/calculation/MeasureHighlightingPanel.tsx
elsaperelli: Good catch!!

zacharyrobin (approved):
🎉💯👌

PR #105, opened 2025-04-21
Small fixes

lmd59:
# Summary
Some useful changes for handling resource viewing. Updates the resource summary to properly use certain characteristics of the primary date path. Update sorting by date to use the date comparison instead of string comparison.

## New Behavior
Resource cards work better for choice type primary paths (i.e. MedicationRequest) and non-CodeableConcept primary paths (i.e. MessageDefinition, OperationDefinition)

## Code Changes

- `dates.ts` - adds a Date-typed `sortDate` field to the returned object in `dateForResource`
- `ResourceDisplay.tsx` - uses the `sortDate` to sort resources by date
- `code.ts` - fully utilizes the path information from the `parsedCodePaths` to handle choice types and alternate code types

# Testing Guidance

- `npm run check`
- `npm run dev`
- Test with a patient with many resources of different dates to make sure date sorting (asc and desc) is done correctly. Create a resource without a date to make sure it is sorted last.
- Test with or create your own MedicationRequest, MessageDefinition, and OperationDefinition resources to make sure that the code from the primary code path are shown for each in the resource info card.

### DIFF
--- a/util/fhir/dates.ts
+++ b/util/fhir/dates.ts
@@ -13,6 +13,12 @@ export interface DateInfo {
   [dateField: string]: DateFieldInfo;
 }
 
+export interface DateFormatInfo {
+  date: string;
+  dateType: string;
+  sortDate?: Date;
+}
+
 /**
  * Parses the primary date path info and populates date info to satisfy date filter specified in the passed-in data requirement
  * @param {Object} resource FHIR resource being created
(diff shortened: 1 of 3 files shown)


util/fhir/dates.ts
```diff
@@ -166,17 +166,18 @@ export function getRandomPeriodInPeriod(start: string, end: string): fhir4.Perio
 }
 
 /**
- * Takes in resource and returns the Date and dateType formatted nicley (used in Resource Display cards)
+ * Takes in resource and returns the Date and dateType formatted nicely (used in Resource Display cards)
  * @param {fhir4.FhirResource} resource resource to have date formatted
  * @param {string} end end date string for the period
  * @returns {Object} two strings  {date, dateType} if date is a period it will be in the format of mm/dd/yyyy-mm/dd/yyyy
```
zacharyrobin, line 172: Small change to update the comment return type to match something like:
`@returns {Object}  {date <Date>, dateType  <string>, sortDate <Date>} if date is a period it will be in the format of mm/dd/yyyy-mm/dd/yyyy. sortDate is initially set to startTime if date is a period`

zacharyrobin (changes requested):
Looks good! thanks for helping clean up the date logic, definitely makes more sense to compare the objects instead of strings. Just one small on a comment so it reflects some of the changes

util/fhir/dates.ts
zacharyrobin: or 
```
/**
 * Takes in resource and returns the Date,  dateType and sortDate formatted nicely (used in Resource Display cards)
 * @param {fhir4.FhirResource} resource resource to have date formatted
 * @param {string} end date string for the period
 * @returns {Object}  {date <Date>, dateType  <string>, sortDate <Date>} if date is a period it will be in the format of mm/dd/yyyy-mm/dd/yyyy. sortDate is initially set to startTime if date is a period
 */
```

lmd59: Thanks for catching! There were a couple of futzy things in here, so that was the bump I needed to make a type for the return object.

zacharyrobin (approved):
Looks great! All checks passed, tested with pt with larger resource and sorting / sorting works as expected. Primary code paths also shown correctly ✅

PR #107, opened 2025-05-09
Rename test cases, select measure report, and default trustMetaProfile

lmd59:
# Summary
Rename test cases, select measure report, and default trustMetaProfile

## New Behavior

- When test cases are uploaded along with a README.txt mapping, the patient resource is renamed to the mapped test case name. 
- During upload, the measure report used to populate the desired populations must have a `measure` field matching the loaded measure url. 
- The trustMetaProfile switch defaults to true.

## Code Changes
- Update import modal to allow for .txt upload
- Update `handleSubmittedImport` in the PatientCreationPanel to create a mapping from the readme.txt
- Update `trustMetaProfile` atom to default true
- Update `bundleToTestCase` in the import util to match the measureReport measure field and to rename the patient resource as available
- update README and tests

# Testing Guidance
- `npm run check`
- `npm run dev`
- Test the new behavior by uploading a measure and test cases from MADiE that include multiple measureReports and a readme.txt


util/import.ts
```diff
@@ -10,18 +11,25 @@ import { TestCaseInfo } from '../state/atoms/patientTestCase';
...
   const desiredPopulations: string[] = [];
-  const testCaseMeasureReportArr = bundle.entry.filter(isTestCaseMeasureReport);
+  const testCaseMeasureReportArr = bundle.entry.filter(
+    e => isTestCaseMeasureReport(e) && (e.resource as MeasureReport).measure === measureURL
+  );
   if (testCaseMeasureReportArr.length > 1) {
     // TODO: Once we have import errors persist on page, replace this!!!
     throw new Error(
-      `Expected 0 or 1 test case measure reports in bundle, but found ${testCaseMeasureReportArr.length}`
+      `Expected 0 or 1 test case measure reports in bundle that match the loaded measure, but found ${testCaseMeasureReportArr.length}`
```
elsaperelli, line 32: We may have to make this message a little clearer. Correct me if I am wrong, but I think we could still have 1 measure report in the bundle (theoretically) but that doesn't match the loaded measure? (I don't think we have actually seen this tho)

elsaperelli (approved):
Looking awesome!!! This is such a great improvement to fqm-testify. One small suggestion (can be ignored) so otherwise approved!

util/import.ts
hossenlopp: My thought is that if there is a single test case MeasureReport in the bundle, we should use it for the expected values. But if there are multiple, then we should find the one that matches the Measure's url.

hossenlopp (changes requested):
One nit pick and one suggestion for how to deal with some of the weird test case MeasureReport scenarios.

elsaperelli (approved):

hossenlopp (approved):

PR #108, opened 2025-05-30
Initial totals and coloring

lmd59:
# Summary
Adds a total across all patients for the population counts screen as well as red/green highlighting for whether a  population value matches the expected value.
There are a couple of small design points on this, but this is probably the most expedient approach. Interested in whether we think this meets the need well.

## New Behavior
When the user shows the table of population results, the table shows totals (first row of the table) and red/green highlighting against desired populations.

## Code Changes

- Changes within the PopulationResultsTable.tsx file to total all results and color individual cells.

# Testing Guidance

- `npm run check`
- `npm run dev`
- Load a measure and test cases, click "Calculate Population Results", and look at the table. Experiment with creating new test cases or loading other test cases.

### DIFF
--- a/components/calculation/PopulationResultsTable.tsx
+++ b/components/calculation/PopulationResultsTable.tsx
@@ -1,8 +1,9 @@
-import { createStyles, Table } from '@mantine/core';
+import { Badge, createStyles, Table } from '@mantine/core';
 import { useMemo } from 'react';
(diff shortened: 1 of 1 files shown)


zacharyrobin (approved):
Tested out the functionality and it works well. Noticed no error. I agree with hoss, maybe we can make a new ticket for the future for having a more clear indication of which lines fail

PR #112, opened 2025-09-08
Real submit data option

lmd59:
# Summary
Add option for the "Submit Data" button to allow either a "proto" submit data (just POST the transaction bundles) or a real submit data that uses the appropriate endpoint.

## New Behavior
"Submit Data" button has two options. The "POST data" option will do the "proto" version, and the "$submit-data" will do the full deqm option.

## Code Changes
- All changes are updates to `PopulationCalculation.tsx` to make the UI change and update the `submitDataToEvaluationService` function to be flexible around which option it's using (both use transaction bundles).

# Testing Guidance
- `npm run check`
- `PORT=3001 npm run dev`
- Start the deqm-test-server local instance (`npm run start`)
- Load a measure that is in the deqm-test-server instance (default ecqm-content-r4-2021 measures, I used CervicalCancerScreening)
- Use http://localhost:3000/4_0_1 for the Evaluation Service URL and click "Verify" then "Next"
- Load some test cases
- Choose the "Submit Data" dropdown from the bottom left and experiment with choosing either option. Observe requests sent in the chrome network tab. The "$submit-data" option should send one fetch to the server, whereas the "POST data" option will do a fetch for each patient.
- To observe a change on the server-side, you can change the id of one of the patients, use the $submit-data option, and observe the patient count on the server going up.

### DIFF
--- a/components/calculation/PopulationCalculation.tsx
+++ b/components/calculation/PopulationCalculation.tsx
@@ -5,12 +5,14 @@ import {
   Drawer,
   Grid,
   Group,
+  Menu,
   Modal,
   Radio,
   Select,
   Space,
   Text,
-  Tooltip
+  Tooltip,
+  useMantineTheme
 } from '@mantine/core';
 import CodeMirror from '@uiw/react-codemirror';
 import { useRecoilValue } from 'recoil';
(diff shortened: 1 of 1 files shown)


elsaperelli (approved):
Beautiful! UI expertise!

